PR #1, opened 2017-12-04
(PDK-492) add pdk components

underscorgan:


Gemfile
```diff
@@ -0,0 +1,17 @@
...
+  else
+    [place, { :require => false }]
+  end
+end
+
+gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '0.14.1')
+gem 'packaging', '~> 0.6', :git => 'https://github.com/puppetlabs/packaging.git'
+gem 'rake', '~> 12.0'
+
+#gem 'rubocop', "~> 0.34.2"
```
mwaggett: Is there a reason this is still here, but commented out?

underscorgan: I copied the gemfile from pdk, that's what they had https://github.com/puppetlabs/pdk-vanagon/blob/master/Gemfile#L17

ext/build_defaults.yaml
```diff
@@ -0,0 +1,28 @@
...
+gpg_key: '7F438280EF8D349F'
+deb_targets: 'trusty-amd64 xenial-amd64'
+rpm_targets: 'el-6-x86_64 el-7-x86_64 sles-11-x86_64 sles-12-x86_64'
+sign_tar: FALSE
+osx_signing_cert: "Developer ID Installer: PUPPET LABS, INC. (VKGLGN2B6Y)"
+osx_signing_keychain: "/Users/jenkins/Library/Keychains/signing.keychain"
+osx_signing_server: 'osx-signer.delivery.puppetlabs.net'
+vanagon_project: TRUE
+apt_repo_name: 'PC1'
+yum_repo_name: 'PC1'
```
mwaggett: Do we want these still?

Gemfile
mwaggett: ookay!

ext/build_defaults.yaml
underscorgan: Copied from PDK. Commit from yesterday was reverted as it broke things, can copy it over when it's fixed. These don't really matter at all though since we aren't generating packages and aren't shipping artifacts publicly

mwaggett: yeah, I was just hoping to avoid having lots of unnecessary things, since I think it's easy to cause confusion.

underscorgan:
@mwaggett removed the repo names

PR #3, opened 2017-12-15
(WIP) Add agent runtime

caseywilliams:
I've started modifying the runtime to be used in puppet-agent, and I'm hoping for some feedback on the direction I'm headed here.

At this point I'm able to build a puppet-runtime package for the agent and use it to build smoke-test-passing puppet-agent packages from the 1.10.x, 5.3.x, and master agent branches for redhat 7. I have not yet tested this on other platforms; A lot more work remains to be done.

This commit has a lot of additions, the bulk of which are resources  copied directly from the puppet-agent packaging repo, but here are the most relevant places for review right now:

agent-runtime project:

- I renamed the pdk runtime project to `pdk-runtime` and added an `agent-runtime` project alongside it.
- The puppet-runtime will only operate out of the master branch, but the puppet-agent repo maintains three (1.10.x, 5.3.x, and master). I've handled this here by setting an `agent_branch` variable in the project config that comes from an `AGENT_BRANCH` environment variable and is used to decide which components to include in the project. Should I instead make three separate runtime project configs?
    - Related: I've been naming the package tarballs based on their git description, but I also included the branch name; Is this valuable? Does anyone have better suggestions about package names?

components:

- The PDK and puppet-agent use near-identical `runtime` components; The only difference is the `RUNTIMEDIR` filepath (`/opt/puppetlabs/(pdk|puppet)/lib`) used in `runtime.sh`. Instead of creating a separate component for the agent's runtime, I used an environment setting in the project to set the project's name for that directory. Is it worth making a separate runtime component for the agent instead?
  - Following from this, would it make sense to rename the `runtime` component to something else? It seems strange to have a component called "runtime" inside a runtime project, but I can't think of a more appropriate name, either.
- The pdk `ruby-stomp` 1.4.4 component is identical to that of the agent; This is good, but the agent's 1.10.x branch uses version 1.3.3; The only differences between the 1.3.3 and 1.4.4. configs are version number and sha, plus the addition of a few patches to 1.3.3. In this commit I kept a single ruby-stomp component and used project settings to decide the version and sha, then used those settings to apply the patches when necessary. I'm not confident about this part and am thinking I'll just create a separate ruby-stomp config; I'm open to suggestions though.

Anything else I haven't considered here?


configs/components/ruby-2.4.2.rb
```diff
@@ -0,0 +1,302 @@
+component "ruby-2.4.2" do |pkg, settings, platform|
+  pkg.version "2.4.2"
+  pkg.md5sum "fe106eed9738c4e03813ab904f8d891c"
+  pkg.url "https://cache.ruby-lang.org/pub/ruby/2.4/ruby-#{pkg.get_version}.tar.gz"
+  pkg.mirror "#{settings[:buildsources_url]}/ruby-#{pkg.get_version}.tar.gz"
+
```
ekinanp: I think you also need to add this part:

https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/ruby-2.4.2.rb#L7-L11

configs/projects/agent-runtime.rb
```diff
@@ -0,0 +1,69 @@
...
+  proj.license "See components"
+  proj.vendor "Puppet, Inc.  <info@puppet.com>"
+  proj.homepage "https://puppet.com"
+
+  proj.setting(:artifactory_url, "https://artifactory.delivery.puppetlabs.net/artifactory")
+  proj.setting(:buildsources_url, "#{proj.artifactory_url}/generic/buildsources")
+
+  # At least on el-7, ruby_bindir should be set to /opt/puppetlabs/puppet/bin/, which is the same as the regular bindir:
+  # This does not seem to arrive here from the agent project's settings.
+  proj.setting(:ruby_bindir, "#{proj.settings[:bindir]}")
```
underscorgan: fyi, it looks like this is only set for windows in the agent config, so that's why it's not making it in for el7

configs/projects/agent-runtime.rb
```diff
@@ -0,0 +1,69 @@
+project 'agent-runtime' do |proj|
+  platform = proj.get_platform
+  # TODO: Is this an acceptable way to inherit from multiple possible agent version branches?
+  proj.setting(:agent_branch, ENV['AGENT_BRANCH'] || 'master')
```
underscorgan: Hm. Originally I was definitely thinking it would be better to have a project per branch, but if there aren't many differences that might be a lot of overhead.

The main complication with this is going to be making sure we have a good naming/versioning scheme so that it's very obvious which runtime artifacts to be pulling in for each of the agent builds. The initial naming scheme is in https://tickets.puppetlabs.com/browse/VANAGON-98, but since you're adding the first / possibly only project with branches maybe it would make more sense to have something like agent-runtime-<branch>-<version>.

configs/projects/agent-runtime.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  proj.inherit_settings 'puppet-agent', 'git://github.com/puppetlabs/puppet-agent', proj.settings[:agent_branch]
+  # Set the version based on `git describe`, then prepend the branch name
+  # TODO: (It seems like there must be a better way to name the package - Is this valuable?)
```
underscorgan: I don't think you want to be using `package_version` here, that will be the version from agent.

I think if we go forward with a single agent project for all the branches you probably want to replace the `proj.version` line below with
`proj.version "#{proj.settings[:agent_branch]}-#{proj.get_version}"`

configs/projects/agent-runtime.rb
```diff
@@ -0,0 +1,69 @@
...
+  proj.generate_archives true
+  proj.generate_packages false
+
+  proj.description "The puppet agent runtime contains third-party components needed for the puppet agent"
+  proj.license "See components"
+  proj.vendor "Puppet, Inc.  <info@puppet.com>"
+  proj.homepage "https://puppet.com"
+
+  proj.setting(:artifactory_url, "https://artifactory.delivery.puppetlabs.net/artifactory")
+  proj.setting(:buildsources_url, "#{proj.artifactory_url}/generic/buildsources")
```
underscorgan: did these not make it over from the agent config?

configs/projects/agent-runtime.rb
underscorgan: But probably a good idea to double check that setting this won't break windows, when you get to working on that platform.

configs/components/ruby-2.4.2.rb
caseywilliams: There are a whole lot of places these components will need to be modified for other platforms; I haven't really begun that part yet. My plan for this work was something along these lines:

- Make the existing components in puppet-runtime work for all agent branches on one platform first
- Add any new components that make sense from the puppet-agent repo (e.g. other ruby gems if possible, etc.)
- Add support for other platforms

configs/projects/agent-runtime.rb
```diff
@@ -0,0 +1,69 @@
...
+  proj.component "libxml2" unless platform.is_windows?
+  proj.component "libxslt" unless platform.is_windows?
+  proj.component "ruby-augeas" unless platform.is_windows?
+  # We only build ruby-selinux for EL 5-7
+  if platform.name =~ /^el-([567])-.*/ || platform.is_fedora?
+    proj.component "ruby-selinux"
+  end
+
+  # Here's an idea around setting disparate versions for components when their configs don't differ very much;
+  # Or is it better to just make separate component config files for each version?
```
underscorgan: I think it's probably better to have separate component configs for each version, I think that's a bit more consistent / straightforward

resources/files/runtime/runtime.sh
```diff
@@ -20,7 +20,11 @@ else
   fi
 fi
 
-RUNTIMEDIR=/opt/puppetlabs/pdk/lib
+if [ -z "$PROJECT_SHORTNAME" ]; then
+  PROJECT_SHORTNAME="pdk"
+fi
+
+RUNTIMEDIR=/opt/puppetlabs/$PROJECT_SHORTNAME/lib
```
underscorgan: This seems reasonable.

configs/projects/agent-runtime.rb
caseywilliams: I think I added this because the agent 1.10.x configs specifically didn't make it - they're done slightly differently from the other branches - seems like this is something we should just fix in the agent if possible.

configs/projects/agent-runtime.rb
caseywilliams: I think I will go ahead and make separate project configs for each branch -- seems like as I add more cases to this file the difficulty of reading it could surpass the convenience of a single config.

underscorgan: That makes sense. If you find you have a lot of duplicated code can always go for something like what we do in pl-build-tools: https://github.com/puppetlabs/pl-build-tools-vanagon/blob/master/configs/projects/pl-cmake.rb#L2-L3

PR #5, opened 2018-01-04
(PA-1765) Add runtime projects for puppet-agent master, 5.3.x, and 1.10.x

caseywilliams:
**What changed**

This commit adds three new runtime projects, one for each of puppet-agent's branches.  The `agent-runtime-master`, `agent-runtime-5.3.x` and `agent-runtime-1.10.x` projects extend a common base project, `base-agent-runtime`, which contains all of the agent's common settings (the acutal differences amongst the branch-specific configs are very minor). 

Also added are several new components and platforms from the agent. In summary:

- `ruby-2.4.2` for use in puppet-agent master and 5.3.x (puppet-agent 1.10.x uses the existing ruby 2.1.9)
- `ruby-stomp-1.3.3` (in addition to the existing 1.4.4)
- all rubygems from the agent; I set these up to extend from a common `base-rubygem` component, since they had a lot in common.
- `libedit`
- lots of platforms
- all associated patch/support files

**Handling differences between projects**

The existing pdk project and the puppet-agent projects require a few components (e.g. runtime, openssl) which are mostly the same, but have configurations that differ to some extent. To handle these cases I added a new `:runtime_project` string setting to all projects. This is set to `pdk` or `agent` and used for conditional settings within existing components or to select a separate component name. Examples:

- `openssl` has been broken into `openssl-agent` and `openssl-pdk` - this component had by far the most differences between projects (flags and versions)
- `runtime` has also been broken into `runtime-agent` and `runtime-pdk` - this difference was more minor (just an environment variable change to adjust the runtime.sh script's target directory).

**Other**

- I uncommented all of the `pkg.replaces` lines in the existing components, since they are present in the puppet-agent configs. Will that impact the pdk negatively? What's a good way to validate that the pdk runtime still works in general after all of these additions?
- I tested agents built with this runtime for master, 5.3.x, and 1.10.x on a subset of the agent platforms; All acceptance tests passed - jobs are here:
    - [master](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/7/)
    - [5.3.x](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/8/)
    - [1.10.x](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/9/)


configs/components/puppet-ca-bundle.rb
```diff
@@ -1,17 +1,39 @@
 component "puppet-ca-bundle" do |pkg, settings, platform|
   pkg.load_from_json("configs/components/puppet-ca-bundle.json")
 
-  pkg.build_requires "openssl"
+  pkg.build_requires "openssl-#{settings[:runtime_project]}"
+
+  java_available = true
```
caseywilliams: I copied this java check from the puppet-agent repo; Is this acceptable to have here, or should I separate this into two different `puppet-ca-bundle` components?

configs/components/curl.rb
```diff
@@ -9,17 +9,16 @@
     pkg.apply_patch 'resources/patches/curl/curl-7.55.1-aix-poll.patch'
   end
 
-  pkg.build_requires "openssl"
+  pkg.build_requires "openssl-#{settings[:runtime_project]}"
```
underscorgan: I'd definitely prefer to have some branching in the openssl component to see if we can combine these into one component, or figure out where we actually need differences vs where the differences are just due to drift. I won't block on this for this PR, but I consider having two components for the same version of the same software sort of a last resort.

configs/components/puppet-ca-bundle.rb
underscorgan: This won't work for PDK, can we just set that variable to 'true' based on which runtime we're building, rather than splitting them apart?

configs/components/ruby-2.4.2.rb
```diff
@@ -0,0 +1,300 @@
...
+    pkg.add_source 'file://resources/files/ruby_242/windows_ruby_gem_wrapper.bat'
+  end
+
+  pkg.replaces 'pe-ruby'
+  pkg.replaces 'pe-ruby-mysql'
+  pkg.replaces 'pe-rubygems'
+  pkg.replaces 'pe-libyaml'
+  pkg.replaces 'pe-libldap'
+  pkg.replaces 'pe-ruby-ldap'
+  pkg.replaces 'pe-rubygem-gem2rpm'
```
underscorgan: I'm going to have to think about this more, if it makes sense to keep these in the runtime or if they should move into the runtime component added to the runtime component in puppet-agent. I'm pretty sure as it is these won't persist into the p-a package

branan:
It makes me kind of nervous that openssl is configured very differently between PDK and Agent. Is there documentation of how and why these differences were decided upon? Has there been an effort to unify this configuration?

/cc @jsane for thoughts on openssl configuration / security implications

caseywilliams:
I've reworked the openssl component to synthesize the pdk and agent versions - I'll be running tests this weekend, but it should be easier to tell where they differ now.

jsane:
I too am curious about what appears to be consciously chosen cipher/feature differences across agent and PDK. Won't mind enabling CMS in PDK assuming it is for a reason but wondered about  SEED. I would have liked/preferred to disable GOST, SRP, RC5 even in agent (not aware of anything requiring these but I will check it/test once).

configs/components/openssl.rb
```diff
@@ -1,88 +1,215 @@
...
-  pkg.url "https://www.openssl.org/source/openssl-#{pkg.get_version}.tar.gz"
+component 'openssl' do |pkg, settings, platform|
+  if settings[:runtime_project] == 'agent'
+    pkg.version '1.0.2k'
+    pkg.md5sum 'f965fc0bf01bf882b31314b61391ae65'
+    pkg.replaces 'pe-openssl'
+  else
+    pkg.version '1.0.2m'
+    pkg.md5sum '10e9e37f492094b9ef296f68f24a7666'
+  end
```
underscorgan: I understand not wanting to introduce too many changes when moving to a runtime, but we should leave a comment in here to get rid of the block when agent is updated to latest openssl (there have been more security things). One of the nice things with the runtime is everything will get these updates together :)

configs/components/ruby-2.1.9.rb
```diff
@@ -5,20 +5,18 @@
...
-  # pkg.replaces 'pe-ruby-ldap'
-  # pkg.replaces 'pe-rubygem-gem2rpm'
+  if settings[:runtime_project] == 'agent'
+    pkg.replaces 'pe-ruby'
+    pkg.replaces 'pe-ruby-mysql'
+    pkg.replaces 'pe-rubygems'
+    pkg.replaces 'pe-libyaml'
+    pkg.replaces 'pe-libldap'
+    pkg.replaces 'pe-ruby-ldap'
+    pkg.replaces 'pe-rubygem-gem2rpm'
```
underscorgan: I think these replaces should move into the agent-runtime component in puppet-agent, from thinking on it more I think that seems like the most straightforward way to manage these.

configs/components/ruby-2.4.2.rb
```diff
@@ -0,0 +1,302 @@
...
+  end
+
+  if settings[:runtime_project] == 'agent'
+    pkg.replaces 'pe-ruby'
+    pkg.replaces 'pe-ruby-mysql'
+    pkg.replaces 'pe-rubygems'
+    pkg.replaces 'pe-libyaml'
+    pkg.replaces 'pe-libldap'
+    pkg.replaces 'pe-ruby-ldap'
+    pkg.replaces 'pe-rubygem-gem2rpm'
```
underscorgan: These should move into the agent-runtime component in puppet-agent

configs/components/ruby-augeas.rb
```diff
@@ -4,7 +4,7 @@
   pkg.url "http://download.augeas.net/ruby/ruby-augeas-#{pkg.get_version}.tgz"
   pkg.mirror "#{settings[:buildsources_url]}/ruby-augeas-#{pkg.get_version}.tar.gz"
 
-  # pkg.replaces 'pe-ruby-augeas'
+  pkg.replaces 'pe-ruby-augeas' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/ruby-selinux.rb
```diff
@@ -13,7 +13,7 @@
     pkg.mirror "#{settings[:buildsources_url]}/libselinux-#{pkg.get_version}.tar.gz"
   end
 
-  # pkg.replaces 'pe-ruby-selinux'
+  pkg.replaces 'pe-ruby-selinux' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/ruby-shadow.rb
```diff
@@ -0,0 +1,66 @@
+component "ruby-shadow" do |pkg, settings, platform|
+  pkg.version "2.3.3"
+  pkg.md5sum "c9fec6b2a18d673322a6d3d83870e122"
+  # I am unable to find ruby-shadow-2.3.3 source anywhere other than our own website. Upstream appears to be dead.
+  pkg.url "https://downloads.puppetlabs.com/enterprise/sources/3.8.3/solaris/11/source/ruby-shadow-#{pkg.get_version}.tar.gz"
+  pkg.mirror "#{settings[:buildsources_url]}/ruby-shadow-#{pkg.get_version}.tar.gz"
+
+  pkg.replaces 'pe-ruby-shadow' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/ruby-stomp-1.3.3.rb
```diff
@@ -0,0 +1,32 @@
+component "ruby-stomp-1.3.3" do |pkg, settings, platform|
+  pkg.version "1.3.3"
+  pkg.md5sum "50a2c1b66982b426d67a83f56f4bc0e2"
+  pkg.url "https://rubygems.org/downloads/stomp-#{pkg.get_version}.gem"
+  pkg.mirror "#{settings[:buildsources_url]}/stomp-#{pkg.get_version}.gem"
+
+  pkg.replaces 'pe-ruby-stomp' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/ruby-stomp-1.4.4.rb
```diff
@@ -1,10 +1,10 @@
-component "ruby-stomp" do |pkg, settings, platform|
+component "ruby-stomp-1.4.4" do |pkg, settings, platform|
   pkg.version "1.4.4"
   pkg.md5sum "1224b9efe7381cea25c506c9c6e28373"
   pkg.url "https://rubygems.org/downloads/stomp-#{pkg.get_version}.gem"
   pkg.mirror "#{settings[:buildsources_url]}/stomp-#{pkg.get_version}.gem"
 
-  # pkg.replaces 'pe-ruby-stomp'
+  pkg.replaces 'pe-ruby-stomp' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/rubygem-deep-merge.rb
```diff
@@ -0,0 +1,13 @@
+component "rubygem-deep-merge" do |pkg, settings, platform|
+  instance_eval File.read('configs/components/base-rubygem.rb')
+  pkg.version "1.0.1"
+  pkg.md5sum "6f30bc4727f1833410f6a508304ab3c1"
+  pkg.url "https://rubygems.org/downloads/deep_merge-#{pkg.get_version}.gem"
+  pkg.mirror "#{settings[:buildsources_url]}/deep_merge-#{pkg.get_version}.gem"
+
+  pkg.replaces "pe-rubygem-deep-merge" if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

configs/components/rubygem-net-ssh.rb
```diff
@@ -0,0 +1,13 @@
+component "rubygem-net-ssh" do |pkg, settings, platform|
+  instance_eval File.read('configs/components/base-rubygem.rb')
+  pkg.version "4.1.0"
+  pkg.md5sum "6af1ff8c42a07b11203058c9b74cbaef"
+  pkg.url "https://rubygems.org/downloads/net-ssh-#{pkg.get_version}.gem"
+  pkg.mirror "#{settings[:buildsources_url]}/net-ssh-#{pkg.get_version}.gem"
+
+  pkg.replaces 'pe-rubygem-net-ssh' if settings[:runtime_project] == 'agent'
```
underscorgan: This should move into the agent-runtime component in puppet-agent

caseywilliams:
Thanks @underscorgan! :) Updated to remove all the `replaces` lines and add a note about bumping openssl

configs/components/puppet-ca-bundle.rb
```diff
@@ -4,14 +4,41 @@
   pkg.build_requires "openssl"
```
geoffnichols: Note: I'm not able to comment on it directly, but it looks like puppet-agent is using version 1.0.8 of the puppet-ca-bundle (in configs/components/puppet-ca-bundle.json) on all three puppet-agent branches.

configs/components/rubygem-net-scp.rb
```diff
@@ -0,0 +1,17 @@
+component "rubygem-net-scp" do |pkg, settings, platform|
+  instance_eval File.read('configs/components/base-rubygem.rb')
```
geoffnichols: Since we're no longer building or shipping huaweios agents, we may be able to drop this and a other few rubygems from the runtime. 

Based on the set of huaweios-specific gems defined in the [puppet-agent project config](https://github.com/puppetlabs/puppet-agent/blob/e5503651654452d62b18a3da08c9ccd2539bd437/configs/projects/puppet-agent.rb#L275-L283), we may possibly be able to drop these from puppet-runtime:

- rubygem-net-ssh-telnet2
- rubygem-net-scp
- rubygem-mini_portile2
- rubygem-pkg-config
- rubygem-net-netconf
- rubygem-nokogiri

/cc @ScottGarman

resources/files/ruby_219/rbconfig/rbconfig-powerpc-ibm-aix5.3.0.0.rb
```diff
@@ -0,0 +1,268 @@
+
```
geoffnichols: We're no longer building/shipping agents for AIX 5.3, so we may be able to drop this file.

resources/files/ruby_242/rbconfig/rbconfig-powerpc-ibm-aix5.3.0.0.rb
```diff
@@ -0,0 +1,273 @@
+
```
geoffnichols: We're no longer building/shipping agents for AIX 5.3, so we may be able to drop this file.

resources/files/ruby_219/rbconfig/rbconfig-aarch64-redhat-linux.rb
```diff
@@ -0,0 +1,268 @@
+
```
geoffnichols: As we're adding things under `resources/files/ruby_*` and `resources/patches/*`, should we also remove the corresponding "source"  files from puppet-agent?

configs/components/ruby-2.1.9.rb
```diff
@@ -5,21 +5,9 @@
   pkg.mirror "#{settings[:buildsources_url]}/ruby-#{pkg.get_version}.tar.gz"
```
ekinanp: There's some shared code here and in ruby-2.4.2, from what I can find here:

https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L7-L9
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L7-L9

Common patches for AIX and Windows

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L127-L130
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L128L131

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L132-L142
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L133-L144

(one has an extra zlib-devel on AIX, aside from that things are the same)

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L144-L150
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L146-L152

Most of these parts:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L152-L275
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L171-L291

If you're interested, I have some ideas on how to clean this stuff up

ekinanp: There's some shared code here and in ruby-2.4.2, from what I can find here. If you're interested, I have some ideas on how to clean this stuff up. 

https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L7-L9
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L7-L9

Common patches for AIX and Windows

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L127-L130
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L128L131

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L132-L142
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L133-L144

(one has an extra zlib-devel on AIX, aside from that things are the same)

This part:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L144-L150
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L146-L152

Most of these parts:
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.1.9.rb#L152-L275
https://github.com/caseywilliams/puppet-runtime/blob/agent-runtime/configs/components/ruby-2.4.2.rb#L171-L291

configs/components/puppet-ca-bundle.rb
caseywilliams: Hmm, this is a conundrum - the puppet-ca-bundle component version is loaded from json in both the runtime (v1.0.7) and the existing agent (v1.0.8) - Does it make sense to have multiple json files for the component and choose one based on the project? @underscorgan do you have ideas?

configs/components/rubygem-net-scp.rb
caseywilliams: We should probably take this opportunity to remove references to huawei throughout the agent - I'll update this and make agent PRs unless anyone objects.

configs/components/ruby-2.1.9.rb
caseywilliams: Please do let me know your suggestions

caseywilliams: Speaking of ruby 2.4.2, it's worth noting in this PR that there's are open PRs to the agent that update ruby 2.4.2 and patch ruby 2.1.9 - I will need to update this once those are merged.

ekinanp: It's something like an instance_eval at some point in the middle or the end since here, you need to have setup things like the `rbconfig_info` + the `base` directory for patches so you can't do it at the beginning like you could w/ the `rubygem` stuff.

In the component files, you would declare the `rbconfig` variable and the custom `base` dir variable, the latter is custom since you will probably be applying some common patches (e.g. for AIX) in the base component file, whereas `rbconfig` is structurally the same. From there, you can put the common stuff in the base component file and then call an instance_eval, where the base component file's like what you have for `rubygem-base`.

Only fault with that is you lose the nice structure that the current component files give you (apply patches, setup environments + requirements, then set up installation command). If you'd like that structure, then you can still do instance_eval but just have methods for those specific steps (`setup_common_environment`) in the base file that you can call when you need in the specialized files (e.g. `setup_base_environment` => apply environment overrides).

configs/components/puppet-ca-bundle.rb
underscorgan: I suspect the difference is just drift from when the bundle was updated in September, we should update the json to have version 1.0.8

caseywilliams: Updated to 1.0.8, thanks both

configs/components/openssl.rb
```diff
@@ -1,88 +1,215 @@
...
+  ###########
+  # CONFIGURE
+  ###########
+
+  # OpenSSL Configure doesn't honor CFLAGS or LDFLAGS as environment variables.
+  # Instead, those should be passed to Configure at the end of its options, as
+  # any unrecognized options are passed straight through to ${CC}. Defining
+  # --libdir ensures that we avoid the multilib (lib/ vs. lib64/) problem,
+  # since configure uses the existence of a lib64 directory to determine
+  # if it should install its own libs into a multilib dir. Yay OpenSSL!
```
geoffnichols: @jsane when you have a chance, would you please review the openssl configuration?

jsane: Yes I had a question earlier in the discussion somewhere. 

Besides, I was also planning to remove 3DES separately (to address  https://www.openssl.org/blog/blog/2016/08/24/sweet32/) at least for agent. After doing some experimentation found that one also need to disable SRP, PSK for 3DES to be completely removed. Since this is going to get tested may I request to also include this change (no-3des no-srp no-psk). thx

caseywilliams:
Okay, I have three new commits here: 

- Both ruby versions now extend a common `base-ruby.rb` component (totally open to reverting this if it seems not-good to others, though)
- Gemfile has been updated to ship to artifactory
- Updated openssl flags (apply no-gost and no-rc5 to all projects, introduce no-3des, no-srp, no-psk) in response to @jsane's feedback (thanks!)

I'm going to run the agent branches built on this runtime through the agent ad-hoc pipeline again this weekend to verify everything still seems to work.

jsane:
@caseywilliams Apologies for my rather premature suggestion on updating the ciphers - particularly on disabling 3des and the supposedly dependent ones of srp, psk. I noticed that (only) windows version of PA builds fail with those flags. Unless you have already seen the same failures, pl disregard adding the : no-3des, no-srp, no-psk flags. Inconvenience regretted.

caseywilliams:
@jsane - indeed, the windows builds failed when I included `no-psk`, but that was the only failure - I left it configured to use `no-3des`, `no-srp`, and `no-psk` everywhere except windows, which only has `no-3des` and `no-srp`. I'm waiting on some acceptance tests to finish running for the agent packages I built with this configuration, so we'll see how it goes, but assuming they pass would you like to keep that set of flags or revert?

jsane:
Right no-psk definitely causes one failure though I had seen another one which did not seem clearly tied to it. If everything is building clean (& passing all tests) we can keep the flags as you have done. However am suspecting it might not disable 3des on windows but that is ok (at least it would maintain status quo). thanks

caseywilliams:
Okay, agent builds with this runtime [seem to be passing acceptance tests](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/49/), but since the opening of this PR there have been a couple component bumps in the agent that need to be migrated here before merge. I'll do that and run a final set of tests - after that happens I believe this is good to go.

caseywilliams:
[Here's an ad-hoc run](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/77/) with some passing acceptance tests for the master branch agent and runtime. I can do some runs of 5.3.x and 1.10.x too if anyone likes.

caseywilliams:
I believe this is really ready to merge now - https://github.com/puppetlabs/puppet-agent/pull/1313 has some passing PR tests for a 1.10.x agent using this runtime.

jsane:
@caseywilliams 
Thanks for all the work/effort. Am not asking for or suggesting any changes but more of a q (asking here for the context). So the base puppet-runtime project's openssl.rb expects any additional project specific settings 
`  # Individual projects may provide their own openssl configure flags:
  project_flags = settings[:openssl_extra_configure_flags] || []
  configure_flags << project_flags << cflags << ldflags`

Where are the additional openssl settings maintained for puppet agent 1.10.x and master (for 5.x) that get munged with the above? The openssl.rb file in former (1.10.x branch of PA) is empty and absent in the later.

branan:
@jsane they're in the various project files in this repo. See, for example, https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/agent-runtime-1.10.x.rb#L8-L17

jsane:
@caseywilliams I recall you mentioning that you had added/kept the no-3des, no-srp and no-psk for all platforms except windows (which would use only first two). Not sure I see them unless am missing again. thx,

caseywilliams:
@jsane #29 restored the openssl configure options to their original state - The new changes mentioned in this thread broke acceptance tests for some modules consuming the agent, and in the interest of getting the runtime changes merged to the agent as seamlessly as possible, we abandoned the openssl configure modifications. At this point, the runtime is present in all branches of puppet-agent, though, so there is more room to experiment now. If you'd like to revisit those changes, I can open a ticket - just let me know what you'd like to see.

PR #6, opened 2018-02-05
Add three new platforms, fix ruby location for pdk

caseywilliams:
There are two commits here:

- Add cumulus, arista, and redhat-fips platforms in support of the agent runtime ([here's an example build](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1378/)), and
- Fix the ruby path for pdk ([example build](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1377/console))- puppet-agent uses the prefix as the ruby installation directory, but pdk uses the `:ruby_dir` setting - this updates the agent runtime to use the prefix directory unless `:ruby_dir` is explicitly set, and uses `:ruby_dir` as the ruby installation directory generally.

### DIFF
--- a/configs/components/base-ruby.rb
+++ b/configs/components/base-ruby.rb
@@ -4,11 +4,11 @@
 
 # Condensed version, e.g. '2.4.3' -> '243'
 ruby_version_condensed = settings[:ruby_version].tr('.', '')
-# Base version, e.g. '2.4.3' -> '2.4'
-ruby_version_short = settings[:ruby_version].gsub(/(\d)\.(\d)\.(\d)/, '\1.\2')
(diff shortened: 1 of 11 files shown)


caseywilliams:
I want to add the win32-eventlog gem as well; Marking don't merge for now.

caseywilliams:
I've found some issues with the augeas bump, so I've removed that from this PR in favor of opening another one once I've resolved them. [Here](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1383/) are some successful builds of the p-a runtime as this PR stands now.

ScottGarman:
👍

PR #7, opened 2018-02-20
(maint) Allow for multiple augeas versions

caseywilliams:
Branches of puppet-agent now variously use three different versions of augeas - 1.4.0, 1.8.1, and 1.10.1. This PR adds support for all three versions (pdk uses version 1.8.1, which is currently the only version of augeas available in puppet-runtime).

Augeas presents a bit of a problem in terms of how to handle multiple possible versions; `ruby-augeas` depends on having _some_ version of augeas, but it doesn't matter which one. Attempting to successfully include a `build_requires 'augeas'` line in the `ruby-augeas` component thus becomes tricky when there are multiple possible augeas versions. If we added separate, version-specific augeas components (e.g. `augeas-1.4.0.rb`), it seems like we'd need to either:
- carry some information about the augeas version into the ruby-augeas component in order to set up the requires relationship correctly, or
- just carefully include the augeas component before ruby-augeas in all cases, hope that no one changes the order, and leave out the `build_requires` in `ruby-augeas`.

Neither of these sound great. I decided to do this:
- Added an `:augeas_version` setting in the projects that need augeas (i.e. all of them)
- Kept a single `augeas.rb` component, which looks to the `:augeas_version` setting for a version if available; If none is supplied, it uses 1.8.1 by default. `ruby-augeas` can continue to `build_requires` this without having to care about the version.

Here are a few example builds for [1.10.x](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1398/), [5.3.x](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1399/), and [master](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1401/).

If this seems okay, I may refactor a few other components along these lines (e.g. ruby-stomp) to reduce the number of version-specific `component-x.y.z.rb` components (and so that people only have to figure out one approach to version selection while they're trying to work on puppet-runtime). If there's some other, cleaner way to accomplish this with vanagon or a different approach someone can think of, that would also be great to hear about.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -3,10 +3,6 @@ component 'augeas' do |pkg, settings, platform|
   pkg.md5sum '623ff89d71a42fab9263365145efdbfa'
   pkg.url "#{settings[:buildsources_url]}/augeas-#{pkg.get_version}.tar.gz"
 
-  if platform.is_sles? && platform.os_version == '10'
-    pkg.apply_patch 'resources/patches/augeas/augeas-1.2.0-fix-services-sles10.patch'
-  end
-
   pkg.build_requires "libxml2"
 
   # Ensure we're building against our own libraries when present
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -1,7 +1,39 @@
 component 'augeas' do |pkg, settings, platform|
-  pkg.version '1.8.1'
-  pkg.md5sum '623ff89d71a42fab9263365145efdbfa'
-  pkg.url "#{settings[:buildsources_url]}/augeas-#{pkg.get_version}.tar.gz"
+  # Projects may define an :augeas_version setting, or we use 1.8.1 by default:
+  version = settings[:augeas_version] || '1.8.1'
+  pkg.version version
+
+  case version
+    when '1.4.0'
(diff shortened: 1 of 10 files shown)


configs/components/augeas.rb
```diff
@@ -1,12 +1,40 @@
 component 'augeas' do |pkg, settings, platform|
-  pkg.version '1.8.1'
-  pkg.md5sum '623ff89d71a42fab9263365145efdbfa'
-  pkg.url "#{settings[:buildsources_url]}/augeas-#{pkg.get_version}.tar.gz"
+  # Projects may define an :augeas_version setting, or we use 1.8.1 by default:
+  version = settings[:augeas_version] || '1.8.1'
+  pkg.version version
 
-  if platform.is_sles? && platform.os_version == '10'
-    pkg.apply_patch 'resources/patches/augeas/augeas-1.2.0-fix-services-sles10.patch'
```
caseywilliams: SLES 10 is no longer supported in puppet-agent, and pdk doesn't need it

underscorgan: can you move this into a separate commit?

ScottGarman:
Seems reasonable. :+1:

caseywilliams:
@underscorgan yep, ready for merge

PR #12, opened 2018-02-22
(maint) Add fast-gettext 1.1.0

caseywilliams:
puppet-agent 1.10.x uses version 1.1.0 of the fast-gettext rubygem - all other projects use version 1.1.2. This PR adds an optional `rubygem_fast_gettext_version` setting that projects can use to select a version explicitly; 1.1.2 is used by default.

### DIFF
--- a/configs/components/rubygem-fast_gettext.rb
+++ b/configs/components/rubygem-fast_gettext.rb
@@ -1,7 +1,18 @@
 component "rubygem-fast_gettext" do |pkg, settings, platform|
   instance_eval File.read('configs/components/base-rubygem.rb')
-  pkg.version "1.1.2"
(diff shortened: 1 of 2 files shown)


caseywilliams:
[Here are some builds](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1407/) using the newly added 1.1.0 version

underscorgan:
I like using a default. When the component is updated, I don't like the idea of needing to update that variable for all the projects that just need the default/latest version.

caseywilliams:
@underscorgan yes, thanks!

caseywilliams:
@underscorgan ok, rebased - thanks!

PR #16, opened 2018-03-09
(PA-1888) Add custom vendor dir for shared rubygems

caseywilliams:
Ported from puppet-agent; Affects ruby 2.4.3 only - [here's the original commit](https://github.com/puppetlabs/puppet-agent/commit/7d87bab1352db02705cbfce561038ad34521fe48)


caseywilliams:
This change has an associated acceptance test in the agent, but Platform OS was thinking of moving it to the runtime later, once the runtime is being used in the agent.

PR #20, opened 2018-03-12
(PA-1888) Add multi_json rubygem

caseywilliams:
Ported from puppet-agent; used in 5.5.x and master agent projects only. [The addition to puppet-agent is here](https://github.com/puppetlabs/puppet-agent/commit/783204701cd0ddc7db52cd08a064e6bde16918ec) - the `:ruby_vendordir` setting in that commit is preserved in the runtime via `inherit_settings` for now, but it's something we'll need to double-check when the directions of settings inheritance is reversed in the near future.

### DIFF
--- a/configs/components/rubygem-multi_json.rb
+++ b/configs/components/rubygem-multi_json.rb
@@ -1,0 +1,15 @@
+component "rubygem-multi_json" do |pkg, settings, platform|
+  instance_eval File.read('configs/components/base-rubygem.rb')
+  pkg.version '1.13.1'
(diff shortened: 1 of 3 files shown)


caseywilliams:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_agent-runtime_runtime-vanagon-packaging_5.5.x/ shows an example build of this branch for several platforms

caseywilliams:
Thanks, fixed the `GEM_HOME` setting - I did a vanagon `inspect` and found that it was being set correctly but if you'd like to see new some example builds just say the word

PR #21, opened 2018-03-13
(PA-1833) Use local puppet-agent settings instead of `inherit_settings`

caseywilliams:
Previously, the puppet-agent runtime projects inherited their settings from puppet-agent using vanagon's `inherit_settings` feature. This PR ports relevant settings from the agent to puppet-runtime so that `inherit_settings` is no longer necessary here.

Also included is a ruby 2.1.9 patch for AIX which I previously forgot to include (and seems necessary for this change).

I've built this branch and used it in a 1.10.x agent which I ran through the ad-hoc agent pipeline  - you can see [passing acceptance tests for the agent build here](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/98/).

Once the changes in this PR are merged, the plan is to apply a corresponding change to puppet-agent, so that settings aren't duplicated across both projects; Instead, puppet-agent will use `inherit_settings` to gather settings from puppet-runtime, applying anything else it needs (wix settings, for example) afterward.

### DIFF
--- a/configs/components/ruby-2.1.9.rb
+++ b/configs/components/ruby-2.1.9.rb
@@ -99,6 +99,7 @@ component "ruby-2.1.9" do |pkg, settings, platform|
     pkg.apply_patch "#{base}/aix_ruby_2.1_libpath_with_opt_dir.patch"
     pkg.apply_patch "#{base}/aix_ruby_2.1_fix_proctitle.patch"
     pkg.apply_patch "#{base}/aix_ruby_2.1_fix_make_test_failure.patch"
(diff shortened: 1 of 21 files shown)


caseywilliams:
I'm experimenting with merging the corresponding agent changes for this PR up through the agent branches and finding a few things to fix as I run agent acceptance tests on the results; I'll mark this don't merge for now.

caseywilliams:
Closing this PR in favor of splitting the PA-1888 commits into their own dedicated PR

PR #22, opened 2018-03-14
(maint) Unify ruby-stomp configs into a single component

caseywilliams:
Use a project setting to select the ruby-stomp version instead of maintaining separate ruby-stomp components for different versions. This makes the ruby-stomp component consistent with the version decision pattern recently introduced for a few other components (e.g. augeas).


caseywilliams:
@smcclellan I agree! Tests are definitely on the roadmap - https://tickets.puppetlabs.com/browse/PA-1529 exists to get the process started - I'd like to get started on that very soon

ScottGarman:
:+1:

PR #30, opened 2018-03-27
(PA-1886) Remove gettext-setup gem

smcclellan:
Since this gem is no longer used at runtime, it should be removed as a
dependency.


geoffnichols:
Are we good to remove gettext-setup from all branches of puppet-agent?

branan:
Unfortunately, I think we need to leave this in the 5.3 and 5.5 branches of puppet-agent (since we've shipped them with this in a public gem path that modules or user gems might access), so it will have to stay in the runtime for as long as those are supported. We should be able to just leave the component file in the repo, and move the dependency from the base agent project file to the ones that need it (5.3 and 5.5).

PR #33, opened 2018-03-29
(WIP) (PA-1824) Add ruby 2.5.0

McdonaldSeanp:


scotje:
Whoever picks this up should be able to change this to inherit a lot of stuff from the https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/_base-ruby.rb file we added.

McdonaldSeanp:
jenkins please test this with agent-runtime-master

PR #34, opened 2018-03-29
(PA-1886) Fix ordering of components in agent projects

caseywilliams:
Builds of rubygems rely on a runtime project setting, which (for agent
projects) isn't set until the base agent project is evaluated. This
commit moves the inclusion of gettext-setup to after this point.


McdonaldSeanp:
👍

PR #35, opened 2018-03-30
(PDK-784) Allow Ruby 2.1.9 to be reintroduced into pdk-runtime

scotje:
Implements the necessary changes for the pdk-vanagon project to be able to specify additional rubies that should be packaged alongside the default Ruby 2.4.3. For now, Ruby 2.1.9 is the only additional ruby supported, but this should generalize fairly easily to new versions if needed.

Hopefully this has no-impact on puppet-agent builds.


scotje:
This is essentially the same code as http://builds.delivery.puppetlabs.net/puppet-runtime/293a74d36698e0659e8b584e00db3112233654c9/artifacts/ just rebased into more logical commits and no longer pulling settings from my pdk-vanagon branch.

caseywilliams:
I ran this through the agent pipelines this weekend - it looks like we have two sets of problems here on the agent side:

- agent tarballs don't build for cross-compiled platforms or x86 solaris (seems like they can't find their rbconfigs)
- pxp-agent acceptance tests are failing with the successful agent builds (less sure about this one)

Aside from those, everything seems okay. I'll look into what it would take to fix these issues today and comment here once I have some more helpful information.

scotje:
Do you have a link to the rbconfig related build failures @caseywilliams ? I can take a look at those.

caseywilliams:
@scotje sure! [Here are all the build failures](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_agent-runtime_runtime-vanagon-packaging_1.10.x/1/).

scotje:
Awesome, thanks!

caseywilliams:
@scotje agent builds are all green now, and it looks like the previous pxp-agent failure is gone, but I'm seeing several new failures in acceptance for platforms that wouldn't build before - [here's an example run in progress](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/135/) - I can continue to look at this, but if you have ideas they're quite welcome of course!

scotje:
Hmmm, interesting, the facter failures seem to be it not able to find a ruby library where it expects it to be, definitely seems like it could be related, I'll investigate more.

ScottGarman:
:+1:

scotje:
Thanks everyone!

PR #38, opened 2018-04-06
(BOLT-427) Add bolt-runtime

MikaelSmith:


configs/projects/bolt-runtime.rb
```diff
@@ -0,0 +1,138 @@
...
+
+  proj.setting(:platform_triple, platform_triple)
+  proj.setting(:host, host)
+
+  proj.setting(:artifactory_url, "https://artifactory.delivery.puppetlabs.net/artifactory")
+  proj.setting(:buildsources_url, "#{proj.artifactory_url}/generic/buildsources")
+
+  # Define default CFLAGS and LDFLAGS for most platforms, and then
+  # tweak or adjust them as needed.
+  proj.setting(:cppflags, "-I#{proj.includedir} -I/opt/pl-build-tools/include")
```
MikaelSmith: There's some refactoring that could be done to bring pdk inline with this and puppet-agent, and extract a `base-ruby` file for handling some of these settings.

configs/components/rubygem-ffi.rb
```diff
@@ -25,13 +26,26 @@
       else
         pkg.md5sum "0b6fd994826952231d285f078cefce32"
       end
+    when '1.9.23'
```
lucywyman: Seems like this should be pulled out into it's own var, since we already seemed to have missed updating it in both places once?

MikaelSmith: I'm not sure what you're referring to. This was intentionally not updated earlier when 1.9.22 was broken. The case statement is specifically about supporting multiple versions - though it's not clear it's needed anymore. I'm leaving it up to @caseywilliams's team to decide what to do about cleaning up this file.

caseywilliams: puppet-agent#1.10.x uses rubygem-ffi version 1.9.14, and I'd prefer attempting to bump that version to 1.9.23 in a separate PR so that we don't need to block this PR on successful agent acceptance tests.

On the other hand, if you're not in a hurry I can try running a branch using rubygem-ffi 1.9.23 with a 1.10.x agent through the puppet-agent ad-hoc pipeline; If it works, we could remove version 1.9.14 as part of this PR.

MikaelSmith: I'd rather keep it separate.

PR #44, opened 2018-04-16
(PA-1983) Separate shared agent components & settings

caseywilliams:
Managing special cases for agent runtimes while sharing a single common
base project provided too many opportunities for error in the order of
inclusion for settings and components. This commit separates the shared
configuration into two separate files - one each for settings and
components - so that settings can be customized in agent projects before
loading the shared components.

Fixes the gem home for rubygem-hocon, which wasn't correctly set on
5.5.x and master due to this problem.

### DIFF
--- a/configs/projects/_shared-agent-components.rb
+++ b/configs/projects/_shared-agent-components.rb
@@ -1,0 +1,60 @@
+# This "project" is designed to be shared by all puppet-agent projects
+# See configs/projects/agent-runtime-<branchname>.rb
+unless defined?(proj)
+  warn('These are components shared by all puppet-agent projects; They cannot be built as a standalone project.')
+  warn('Please choose one of the other puppet-agent projects instead.')
+  exit 1
+end
+
+########
+# Common components for all versions of puppet-agent
+########
+
+# Common components required by all agent branches
+proj.component 'runtime-agent'
+
+unless proj.settings[:system_openssl]
+  proj.component 'openssl'
+end
+
+proj.component 'curl'
+proj.component 'puppet-ca-bundle'
+proj.component "ruby-#{proj.ruby_version}"
+proj.component 'augeas' unless platform.is_windows?
+proj.component 'libxml2' unless platform.is_windows?
+proj.component 'libxslt' unless platform.is_windows?
+proj.component 'ruby-augeas' unless platform.is_windows?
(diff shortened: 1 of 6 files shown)


caseywilliams:
The vanagon `inspect` outputs for the agent projects look identical to me after this change (barring some reordering), but I've started some test builds just in case; I'll update here once they're finished.

ekinanp:
Hmm, why can't we put the shared settings and components in the base agent project?

caseywilliams:
For PA-1983 specifically, the problem looked like this:

- rubygem-hocon was included in the base project, here: https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/base-agent-runtime.rb#L233 - so it reached all versions of puppet-agent
- We wanted rubygem-hocon to use the shared puppet vendor dir, just like hiera-eyaml et al., but only in 5.5.x and master
- The setting that provides the shared vendor dir is here: https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/agent-runtime-5.5.x.rb#L9 - that setting comes after the load of rubygem-hocon in the base project, so rubygem-hocon never picks up the setting at all, falling back to the default gem directory
- We also can't move the shared vendor dir setting above the inclusion of the base runtime projects, since it depends on a variable (`proj.libdir`) set in that file.

We could have just included rubygem-hocon separately in each project, but that would have been more duplication.

In general, I also think separating the settings and components into different files makes it more obvious where to make changes in the future, especially when those settings need to affect only certain branches of the agent.

ekinanp:
Oh ok, I forgot that `proj.component <component>` goes ahead and evals the `component.rb` file . It seems like it'd be better if the components were loaded after they were all declared, but that's a Vanagon change. Thanks for clarifying!

caseywilliams:
Successfully built a few packages [here](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_agent-runtime_runtime-vanagon-packaging_1.10.x/1/) (1.10.x) and [here](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_agent-runtime_runtime-vanagon-packaging_master/5/) (master) - if anyone would like to see more builds or tests, just let me know.

branan:
Yeah, re-doing Vanagon to be lazy might be nice, but it's a pretty big effort. I'm :+1: on this

PR #45, opened 2018-04-17
(PA-1918) Remove ruby-stomp as a component from agent-runtime-master

ekinanp:
We are deprecating mcollective from the puppet-agent for Platform
6. This includes deprecating ruby-stomp, because mcollective is the
only agent component that depends on it. Thus, work in this commit
removes ruby-stomp as a shared agent component, only declaring it as
a component in the 1.10.x, 5.3.x, and 5.5.x agent branches.

### DIFF
--- a/configs/projects/_shared-agent-components.rb
+++ b/configs/projects/_shared-agent-components.rb
@@ -25,7 +25,6 @@ proj.component 'libxml2' unless platform.is_windows?
 proj.component 'libxslt' unless platform.is_windows?
 proj.component 'ruby-augeas' unless platform.is_windows?
 proj.component 'ruby-shadow' unless platform.is_aix? || platform.is_windows?
-proj.component 'ruby-stomp'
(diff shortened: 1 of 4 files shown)


ekinanp:
Blocked on https://github.com/puppetlabs/puppet-agent/pull/1399

ekinanp:
This is the only change I could find. I also did some grepping of "mco", "marionette", "collective"; nothing else relevant popped up. Also looking at e.g. https://github.com/puppetlabs/puppet-agent/blob/5.4.0/configs/components/marionette-collective.rb (a version of mcollective.rb before we moved over to runtime), stomp is the only component other than ruby that's used by MCO.

branan:
The related PA PR is merged - I believe this can be unblocked now?

ekinanp:
@branan Yup, unblocked!

McdonaldSeanp:
👍

PR #46, opened 2018-04-20
(maint) Use alternate mirror for Solaris 10 builds

geoffnichols:
While opencsw.org's mirror service is unavailable, use alternate
mirror for Solaris 10 dependency packages.


caseywilliams:
jenkins please test this with puppet-agent#1.10.x,puppet-agent#5.5.x on solaris10-64,solaris10-SPARC

geoffnichols:
Analog to https://github.com/puppetlabs/puppet-agent/pull/1406.

ScottGarman:
This may un-block our builds, but beaker also uses OpenCSW for Solaris 10 to install packages related to acceptance testing, so I think we're still likely to run into failures at that step.

ScottGarman:
:+1:

PR #51, opened 2018-05-02
(PA-1881)Update semantic-puppet to latest 1.0.x version

jtappa:


joshcooper:
I think artifactory will transparently fetch the gem the first time it's requested so no need to mirror. Also I think we're also not using artifactory yet (never closer).

caseywilliams:
(agent CI is back on artifactory now)

joshcooper:
@caseywilliams are any steps required to promote this into puppet-agent, or does that happen automagically?

caseywilliams:
@joshcooper the [puppet-runtime pipelines](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-runtime/) will build this revision for all pltaforms and projects, ship the results, and then promote into puppet-agent (and others) once everything is finished

PR #52, opened 2018-05-07
(PDK-915) Migrate PDK Ruby-related settings to pdk-runtime config

scotje:


scotje:
@caseywilliams This is the first PR in the process of switching the PDK settings inheritance around so that we can update Ruby versions without breaking builds. :)

caseywilliams:
jenkins please test this with pdk

scotje:
🎉

PR #56, opened 2018-05-16
(PA-2056) Add these gems to default to shared gem dir

jtappa:


branan:
It'd be nice to have a ticket for this linked to the Server ticket to remove them from that project, so that we know the status of the overall migration etc.

jtappa:
Rather than create a new ticket, seemed like adding PA-1077 made more sense, since this is a step towards consolidating gem dependency declarations.

PR #57, opened 2018-05-17
(PA-1995) Publish yaml settings on build for all projects

caseywilliams:
Writes a yaml file containing the settings used during a build (plus a sha1sum for that file) to the output directory alongside other artifacts.

This is intended to make building downstream projects locally with development builds of puppet-runtime easier. It depends on https://github.com/puppetlabs/vanagon/pull/566, so I'm marking this `do not merge` for now.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ def vanagon_location_for(place)
 end
 
 gem 'artifactory'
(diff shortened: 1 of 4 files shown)


caseywilliams:
The vanagon changes supporting this functionality have been merged, so I'm unblocking this -- I've made all the projects publish their settings as yaml, since it has no effect on the actual contents of the tarballs.

PR #58, opened 2018-05-18
(maint) notice when pkgutil can't find a source mirror

speedofdark:
Solaris10 uses the Vanagon provision_with capability to run a moderately
complex shell fragment to provision things. The issue is that if given a
series of shell commands to execute, only the exit status of the last command
run is noticed when determining success.

We have had issues with the mirror server for Solaris10 dropping off the
net, causing build fails. This change provides the ability to fail when
pkgutil fails to find the mirror, which means the build will fail much faster.


caseywilliams:
We should probably get this into the puppet-agent solaris platform definitions as well (will need a separate PR)

PR #61, opened 2018-05-30
(PA-2064) Add Vanagon inspect checking in Travis CI

MikaelSmith:


MikaelSmith:
Needs TravisCI turned on for this repo to test it.

MikaelSmith:
Passes now. The log is pretty verbose. We could try to do something about that.

caseywilliams:
Hmm, yeah -- [this comment](https://github.com/travis-ci/travis-ci/issues/2285#issuecomment-42724719) suggests that it's possible to tell travis where to fold outputs -- might be worth a try?

MikaelSmith:
Could also use jq to pull out specific sections. I'll try the folding though.

MikaelSmith:
Folding doesn't seem to want to work for more than a few entries.

caseywilliams:
Hmm, alright, thanks for testing it -- I'm ready to merge this as-is unless you want to pursue filtering with jq

MikaelSmith:
Made the output more compact, we just list the components in each project.

caseywilliams:
Great -- LGTM. As an aside, we need to clear up the deprecation warnings for agent --  it'd be great to be able to check for those in this task. I may look into that later.

PR #63, opened 2018-05-31
(maint) Unpin FFI gem in Agent 1.10.x

branan:
The FFI gem was pinned due to some compatibility issues in intervening
versions. Since the latest version appears to work in newer agent
releases, we are unpinning it here in 1.10.x


branan:
Jenkins please test this

PR #68, opened 2018-06-08
(CPR-391) Add alternate Fedora 26, 27 platforms

caseywilliams:
CPR-391 has updated repositories consuming runtime projects so that they
no longer require the `f` prefix on Fedora platform names, but only in
master branches. This change adds new platform definitions for Fedora 26
and 27 without the prefix, but retains the old definitions, since
earlier branches of puppet-agent still require them.

### DIFF
--- a/configs/platforms/fedora-26-x86_64.rb
+++ b/configs/platforms/fedora-26-x86_64.rb
@@ -1,10 +1,11 @@
-platform "fedora-f27-x86_64" do |plat|
+platform "fedora-26-x86_64" do |plat|
   plat.servicedir "/usr/lib/systemd/system"
   plat.defaultdir "/etc/sysconfig"
   plat.servicetype "systemd"
 
   plat.provision_with "/usr/bin/dnf install -y --best --allowerasing autoconf automake rsync gcc make rpmdevtools rpm-libs"
-  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/yum/pl-build-tools-release-#{plat.get_os_name}-27.noarch.rpm"
+  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/yum/pl-build-tools-release-#{plat.get_os_name}-26.noarch.rpm"
   plat.install_build_dependencies_with "/usr/bin/dnf install -y --best --allowerasing"
-  plat.vmpooler_template "fedora-27-x86_64"
+  plat.vmpooler_template "fedora-26-x86_64"
+  plat.dist "fc26"
 end
--- a/configs/projects/agent-runtime-1.10.x.rb
+++ b/configs/projects/agent-runtime-1.10.x.rb
@@ -7,6 +7,17 @@ project 'agent-runtime-1.10.x' do |proj|
   proj.setting :rubygem_net_ssh_version, '4.1.0'
   proj.setting :rubygem_semantic_puppet_version, '0.1.2'
 
+  platform = proj.get_platform
+
+  # This modification to the platform definition is for compatibility with
+  # pre-puppet6 puppet-agent packaging. Fedora platforms for PC1 have have an
+  # `f` prefix before their version numbers and do not use an explicit .dist
+  # string.
+  if platform.name =~ /^fedora-([\d]+)/
+    platform.instance_variable_set(:@name, platform.name.sub(/\d+/, 'f\\0'))
+    platform.instance_variable_set(:@dist, nil)
+  end
+
   ########
   # Load shared agent settings
   ########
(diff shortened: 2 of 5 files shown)


caseywilliams:
jenkins please test this on fedora26-64a,fedora27-64a

configs/platforms/fedora-26-x86_64.rb
```diff
@@ -0,0 +1,11 @@
+platform "fedora-26-x86_64" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
```
ekinanp: Couldn't you just do something like
```
plat.instance_variable_set(:@name, 'fedora-26-x86_64')
```
in the old platform.rb files?

caseywilliams: Does that allow us to make the platform name depend on the project being built? Agent projects older than master still need the `f`.

mwaggett: I think I like this approach better, but I'm curious if you have reasons for your suggestion @ekinanp

ekinanp: I think so, you should be able to do that in the project files, something like:
```
if plat.name =~ /fedora-(\d+)/
  plat.instance_variable_set(:@name, fedora-f\1-x86_64) 
end
```
in e.g. agent-runtime-1.10.x, agent-runtime-5.3.x.

@mwaggett Just seems overkill adding new platform files for something like that is all. I don't have a strong opinion on either approach though.

mwaggett: In my opinion, this seems easier and cleaner than your conditional. I'd love for all the `f` conditionals to be gone asap. When we EOL puppet5 (which I know will be a while), it seems easiest if we could just EOL the f-containing platform configs like any other platform.

ekinanp: Yeah, when you think about it as a separate platform it makes sense. Then my only other suggestion is adding comments in the plat. files indicating why we have things set-up the way we do (e.g. in the'f' prefix ones, something like "We need this for older agent projects, we're planning on removing it once we EOL puppet5" would be useful).

caseywilliams: This approach seems to work in cursory local tests, but I want to make sure they produce expected results everywhere. I think I'm going to build and ship packages for the updated platforms ad-hoc, so that other projects can consume them now, and wait for PR tests to finish (then build pdk and agents) for these changes.

caseywilliams:
jenkins please test this with puppet-agent#master,puppet-agent#5.5.x,puppet-agent#5.3.x,puppet-agent#1.10.x,pdk on fedora26-64a,fedora27-64a

configs/platforms/fedora-26-x86_64.rb
ekinanp: @mwaggett  Seeing the changes that Casey just pushed up, I think putting it in the project files is better (if it works). The main reason is b/c fedora-f27 and fedora-27 are not meant to be different platforms, they are the same platform - fedora-27. The reason we need the 'f' prefix in the former is because older agent projects require it. That's a project-induced restriction. Hence, consolidating the 'f' prefix should belong in the project files.

configs/projects/agent-runtime-1.10.x.rb
```diff
@@ -7,6 +7,17 @@
   proj.setting :rubygem_net_ssh_version, '4.1.0'
   proj.setting :rubygem_semantic_puppet_version, '0.1.2'
 
+  platform = proj.get_platform
+
+  # This modification to the platform definition is for compatibility with
+  # pre-puppet6 puppet-agent packaging. Fedora 26 and 27 must have have an `f`
+  # prefix before their version numbers and do not use an explicit .dist
+  # string.
+  if platform.name =~ /^fedora-2(6|7)/
```
mwaggett: Do you want this limited to 26 and 27? iirc, we're adding fedora 28 rn and that will need to keep the f for everything pre-puppet6 if we want to maintain consistency in the other streams.

caseywilliams: Ah, that makes sense -- I'll update this, thanks

caseywilliams:
A drawback to this approach is that target platforms for CI in pre-puppet6 puppet-agent still include the `f` prefix. This might require maintaining separate platform lists for runtime and agent in those branches.

McdonaldSeanp:
@caseywilliams we already maintain separate lists in CI for each branch, so it shouldn't be a huge deal to maintain them this way.

Additionally once we EOL 1.10 and 5.3 there will only be one branch to worry about.

caseywilliams:
Just updated to improve the comments a bit -- I think this is ready for merge

ScottGarman:
:+1:

PR #70, opened 2018-06-13
(CPR-391) Remove fedora `f` for agent runtime 5.5.x

caseywilliams:
We won't be keeping the `f` prefix for the fedora 28 platform in
puppet-agent 5.5.x -- this restricts the workaround in that project to
Fedora 26 and 27 only.

### DIFF
--- a/configs/projects/agent-runtime-5.5.x.rb
+++ b/configs/projects/agent-runtime-5.5.x.rb
@@ -5,17 +5,6 @@ project 'agent-runtime-5.5.x' do |proj|
   proj.setting :rubygem_net_ssh, '4.1.0'
   proj.setting :rubygem_semantic_puppet_version, '0.1.2'
 
-  platform = proj.get_platform
-
-  # This modification to the platform definition is for compatibility with
-  # pre-puppet6 puppet-agent packaging. Fedora platforms for puppet5 must have
-  # have an `f` prefix before their version numbers and do not use an explicit
-  # .dist string.
-  if platform.name =~ /^fedora-([\d]+)/
-    platform.instance_variable_set(:@name, platform.name.sub(/\d+/, 'f\\0'))
-    platform.instance_variable_set(:@dist, nil)
-  end
-
   ########
   # Load shared agent settings
   ########


configs/projects/agent-runtime-5.5.x.rb
```diff
@@ -5,13 +5,11 @@
...
-
   # This modification to the platform definition is for compatibility with
-  # pre-puppet6 puppet-agent packaging. Fedora platforms for puppet5 must have
-  # have an `f` prefix before their version numbers and do not use an explicit
-  # .dist string.
-  if platform.name =~ /^fedora-([\d]+)/
+  # pre-puppet6 puppet-agent packaging. Fedora 26 and 27 for puppet-agent 5.5.x
+  # must have an `f` prefix before their version numbers and do not use an
+  # explicit .dist string.
+  if platform.name =~ /^fedora-2(6|7)/
```
speedofdark: Dont we need 28 if its for 5.5.x?

speedofdark: Or is that where we drop the f?

caseywilliams: This change removes the workaround for fedora 28 -- the fedora 26 and 27 platforms will still have an `f` appended dynamically during builds because of this chunk of code, but we won't be doing that for fedora 28

speedofdark:
Got it. Thanks.

configs/projects/agent-runtime-5.5.x.rb
mwaggett: so I was just talking with @branan - I think if you update all your fedora platform configs to not have the f, the only difference should be the dist tag, since RE has manually been moving things into non-f directories, e.g. http://yum.puppetlabs.com/puppet5/fedora/27/x86_64/

caseywilliams:
I'm tabling this until after we make some changes around the paths we ship to for internal builds -- labeling as do not merge for now.

caseywilliams:
jenkins please test this with puppet-agent#5.5.x on fedora26-64a,fedora27-64a,fedora28-64a

caseywilliams:
Something is amiss with PR kickoff jobs, which I'll look into, but I've built experimental packages [here](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_agent-runtime_runtime-vanagon-packaging_5.5.x/2/) for fedora 26, 27, and 28

PR #71, opened 2018-06-18
(PDK-1008) Update pdk-runtime to use shared components

bmjen:
This PR is to transition PDK to using the shared rubygem dependencies that are available in puppet-runtime. There will be a follow-up PR to pdk-vanagon to remove the respective components there.


bmjen:
This PR was tested and built successfully here: http://builds.puppetlabs.lan/puppet-runtime/738ec50affe5d6cce34341f435ee82a2383ca8e6/artifacts/?C=M&O=D

PR #77, opened 2018-07-12
(PA-2055) Re-add boost/yaml-cpp, fix windows builds of boost

McdonaldSeanp:
This commit 'reverts the revert' of the original vendored yaml-cpp/boost and updates the windows build of boost to (hopefully) correctly work with Leatherman.

### DIFF
--- a/configs/components/boost.rb
+++ b/configs/components/boost.rb
@@ -104,8 +104,13 @@ component "boost" do |pkg, settings, platform|
 
     gpp = "C:/tools/mingw#{arch}/bin/g++"
 
+    # The default build style used by jam is 'minimal', which builds both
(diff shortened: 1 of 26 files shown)


configs/components/yaml-cpp.rb
```diff
@@ -0,0 +1,96 @@
+component "yaml-cpp" do |pkg, settings, platform|
```
stahnma: Why not just use system yaml-cpp?

configs/components/boost.rb
```diff
@@ -0,0 +1,183 @@
+component "boost" do |pkg, settings, platform|
```
stahnma: Why not use system packages instead of building our own? How does this improve things of pl-boost?

PR #79, opened 2018-07-23
(maint) Reduce choco build log clutter 

Iristyle:
- Add the --no-progress switch in an effort to shrink the size of
   build logs. Choco describes the feature as:

   Do Not Show Progress - Do not show download progress percentages.

   Currently, build logs include extremely verbose output when choco
   packages are downloading, adding many lines per each 1% of a given
   package download. This commit removes that extraneous output.


jcoconnor:
@Iristyle Does this need merged up to ```master``` and other branches (e.g. ```5.3.x``` and 5.5.x```) ?

Iristyle:
Thanks @jcoconnor - this repo doesn't have those branches, so I think we're good!

PR #80, opened 2018-07-26
(BOLT-720) Include common R10k dependency for bolt-runtime

MikaelSmith:
Bolt started shipping R10k, which requires several extra gems.
gettext-setup is the only one that's a common dependency here so far.


MikaelSmith:
Repushed with `R10k` as the canonical spelling.

PR #81, opened 2018-07-26
(maint) Only use gem releases of packaging

MikaelSmith:
Pin to latest gem release of packaging rather than following the 1.0.x
dev branch.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-def vanagon_location_for(place)
+def location_for(place)
   if place =~ /^(git[:@][^#]*)#(.*)/
     [{ :git => $1, :branch => $2, :require => false }]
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -12,7 +12,7 @@ end
 
 gem 'artifactory'
 gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '0.15.12')
-gem 'packaging', :github => 'puppetlabs/packaging', branch: '1.0.x'
+gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99.7')
```
mwaggett: You will need to update the `vanagon_location_for` method above to work with this. There should be prior-art in other repos.
Also, while 0.99.7 is fine I think, I'd use 0.99.8 since that's the latest.

MikaelSmith: Fixed.

PR #83, opened 2018-07-27
(PE-23817) Add pe-installer-runtime

highb:
Add the pe-installer-runtime, based off the bolt-runtime.

### DIFF
--- a/configs/components/runtime-pe-installer.rb
+++ b/configs/components/runtime-pe-installer.rb
@@ -1,0 +1,28 @@
+# This component exists to link in the gcc runtime libraries.
+component "runtime-pe-installer" do |pkg, settings, platform|
+  pkg.environment "PROJECT_SHORTNAME", "installer"
+
+  if platform.is_windows?
+    lib_type = platform.architecture == "x64" ? "seh" : "sjlj"
+    pkg.install_file "#{settings[:gcc_bindir]}/libgcc_s_#{lib_type}-1.dll", "#{settings[:bindir]}/libgcc_s_#{lib_type}-1.dll"
+
+    # zlib, gdbm, yaml-cpp and iconv are all runtime dependancies of ruby, and their libraries need to exist inside our vendored ruby
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
+    pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:ruby_bindir]}/zlib1.dll"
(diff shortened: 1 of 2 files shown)


configs/components/runtime-pe-installer.rb
```diff
@@ -0,0 +1,28 @@
+# This component exists to link in the gcc runtime libraries.
+component "runtime-pe-installer" do |pkg, settings, platform|
```
caseywilliams: Missing `c` here

configs/projects/pe-installer-runtime.rb
```diff
@@ -0,0 +1,142 @@
...
+  proj.component 'rubygem-win32-dir'
+  proj.component 'rubygem-win32-process'
+  proj.component 'rubygem-win32-security'
+  proj.component 'rubygem-win32-service'
+
+  # What to include in package?
+  proj.directory proj.prefix
+
+  # Export the settings for the current project and platform as yaml during builds
+#  proj.publish_yaml_settings
```
caseywilliams: This line allows downstream projects that consume the tarballs produced by puppet-runtime to [load their settings out of the yaml file](https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L15) instead of cloning the entire runtime project to gather the settings that way. You'll may want to uncomment this line even if you don't end up doing inheritance the way puppet-agent does it, since it can be helpful to see the settings for an existing build without having to use vanagon's `inspect` command on the same revision.

highb: Ah, ok. I was just noticing that it errored out when I tried it. I'll try again and post the output here.

configs/components/runtime-pe-installer.rb
highb: I'm confused; where is the missing `c`?

caseywilliams: Hmm, I'm seeing `omponent` instead of `component` in the github ui, but I see `component` the email I got with your comment :woman_shrugging:

highb: Huh, must be a ghost commit that I force pushed over. 👻

configs/projects/pe-installer-runtime.rb
caseywilliams: Ah, @highb one more thing to note about this line - it requires vanagon >= 0.15.12, so if your local version is older than that, it will break the build

highb: A `bundle update` fixed it completely. Thanks!

cthorn42:
👍

PR #84, opened 2018-07-27
(maint) Move httpclient dependency to bolt-runtime

MikaelSmith:
Now that `httpclient` is shipping in puppet-agent's runtime, we can
share the dependency with bolt-runtime.

### DIFF
--- a/configs/projects/bolt-runtime.rb
+++ b/configs/projects/bolt-runtime.rb
@@ -125,6 +125,7 @@ project 'bolt-runtime' do |proj|
   proj.component 'rubygem-minitar'
   proj.component 'rubygem-multi_json'
   proj.component 'rubygem-net-ssh'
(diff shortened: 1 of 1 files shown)


configs/projects/bolt-runtime.rb
```diff
@@ -125,6 +125,7 @@
   proj.component 'rubygem-minitar'
   proj.component 'rubygem-multi_json'
   proj.component 'rubygem-net-ssh'
+  proj.component 'rubygem-httpclient'
```
lucywyman: Awesome! Assuming this means we can remove it from the bolt-vanagon repo?

MikaelSmith: Yes, once this is merged and promoted. It's part of puppet-runtime builds for puppet-agent#master, so we can rely on it for bolt-server as well.

PR #90, opened 2018-08-14
(PE-23817) Update OpenSSL for pe-installer-runtime

highb:


configs/projects/pe-installer-runtime.rb
```diff
@@ -95,7 +96,7 @@
     'no-dtls1',
     'no-idea',
     'no-seed',
-    'no-ssl2-method',
+    # 'no-ssl2-method',
```
branan: should this be removed entirely?

highb: Maybe? I was just copy-pastaing from the pdk-runtime. 🤷‍♂️

branan:
Do you need to branch the runtime for pre-johnson and johnson separately, or are you comfortable with older installers ending up with the updated openssl?

highb:
I'm not even sure we'll start using this runtime in Johnson, and I'm pretty sure we won't be back porting it.

highb:
@branan Removed the commented out openssl option.

branan:
:+1: wasn't sure where/how it was being used. If it's still all in-development it doesn't matter to me :)

PR #100, opened 2018-09-23
(maint) Add patch for windows generated batch files to R2.5

bmjen:
I had to generate the new patch for Ruby 2.5.1, the batch files generated previously weren't working for PDK and errored trying to find the path.

Please let me know if this impacts Agent or Bolt, and I can conditionalize it for PDK only.


MikaelSmith:
Fixing the patch for Ruby 2.5 makes sense. Presumably line numbers changed so it didn't apply correctly.

bmjen:
@MikaelSmith the patch was never there for ruby 2.5, I had to generate it.

bmjen:
@McdonaldSeanp The text for the patch is copy pasta from the other patches, but this impacts packaging of vendored ruby as well. Without the patch, the way the package vendors ruby via vanagon with puppet-runtime, the default batch script ignores the path and ruby is unable to find the bundle executable. This is at least what I've seen on the PDK package without this patch, I am unsure if puppet-agent is impacted as well.

McdonaldSeanp:
@bmjen I totally understand. We can't change this in the 5.x of the agent series (ruby < 2.5) because it was already there, but I would like to investigate why these commands wouldn't work in PDK for 6.0.x (ruby > 2.5).

From the agent perspective I would prefer if we could go so far as to remove this patch from the other versions of ruby too, but that's a breaking change we can't violate.

I'm concerned that there are safety implications of allowing these batch files to load an arbitrary ruby installation. If you ran puppet and the batch file loaded an older version of ruby you could risk having puppet load with vulnerabilities in  the old ruby or one of the gems.

bmjen:
@McdonaldSeanp sure, we were able to work around not having this patch in the PDK project, so we can wait on this PR until we figure out if the general consensus is to not patch these batch files anymore.

I'd be interested in hearing why the puppet 6 agent builds and CI weren't impacted by the lack of this patch.

MikaelSmith:
This looks like it might be related to https://tickets.puppetlabs.com/browse/BOLT-886.

resources/patches/ruby_251/windows_ruby_2.5_fixup_generated_batch_files.patch
```diff
@@ -0,0 +1,86 @@
...
++++ b/lib/rubygems/installer.rb
+@@ -754,14 +754,19 @@ def app_script_text(bin_file_name)
+ 
+   def windows_stub_script(bindir, bin_file_name)
+     ruby = Gem.ruby.gsub(/^\"|\"$/, "").tr(File::SEPARATOR, "\\")
+-    return <<-TEXT
++    return <<-SCRIPT
+ @ECHO OFF
+ IF NOT "%~f0" == "~f0" GOTO :WinNT
+-@"#{ruby}" "#{File.join(bindir, bin_file_name)}" %1 %2 %3 %4 %5 %6 %7 %8 %9
```
MikaelSmith: @McdonaldSeanp this line assumes that you're running `gem install` on an installed version of Ruby, and that the location of that Ruby executable will never change. It uses the absolute path to ruby as returned by `Gem.ruby`.

In our packaging, where we install gems and package them up in a relocatable MSI, we can't assume the absolute path of the Ruby location.

McdonaldSeanp: @MikaelSmith Gem.ruby will return the correctly qualified path to the installed ruby. `Gem.ruby` just uses rbconfig values to identify where ruby is installed.

Rbconfig uses the following two lines to identify the prefix for where ruby is installed:
```
TOPDIR = File.dirname(__FILE__).chomp!("/lib/ruby/2.4.0/x64-mingw32")
```
and
```
CONFIG["prefix"] = (TOPDIR || DESTDIR + "/mingw64")
```
since TOPDIR dynamically parses using the location of the RBConfig file itself, it will return the path to where ruby is actually installed.

So Gem.ruby will return the correct path to ruby.

MikaelSmith: During packaging yes. But then when we package it and put it in an MSI, the location will be wrong.

McdonaldSeanp: We have gem installation tests in the agent that should identify whether or not gem installations are working, so if that was a problem we should have seen some failing tests.

MikaelSmith:
Is there a PDK ticket for this? I believe this is a solution for PA-2203 and BOLT-886.

resources/patches/ruby_251/windows_ruby_2.5_fixup_generated_batch_files.patch
McdonaldSeanp: Nope, nevermind, you referenced this was during the build which makes all of my points useless.

resources/patches/ruby_251/windows_ruby_2.5_fixup_generated_batch_files.patch
```diff
@@ -0,0 +1,86 @@
...
++ECHO.This version of Ruby has not been built with support for Windows 95/98/Me.
+ GOTO :EOF
+ :WinNT
+-@"#{ruby}" "%~dpn0" %*
+-TEXT
++IF EXIST "%~dp0ruby.exe" (
++  SET RUBY_EXE_PATH="%~dp0ruby.exe"
++) ELSE (
++  SET RUBY_EXE_PATH="ruby.exe"
```
branan: I'd really prefer to remove this ELSE case and update the description in the patch to match why we actually need it now, instead of just taking this as-is from what was done in the past. There should be no reason we'd ever need to fall back to another ruby here, and I agree with @McdonaldSeanp that it could potentially cause problems.

If PDK relies on this bit of the behavior, I'd like to better understand why/how so we can maybe not do that.

Otherwise I think this is fine to land

McdonaldSeanp: Yeah, that would work for me.

MikaelSmith: Yeah, should work.

MikaelSmith:
Looks like the patch didn't apply correctly.

bmjen:
Doh, @MikaelSmith I was test building the patch still. I mentioned it in hipchat, should have done so here as well. I'll generate a new patch and PR.

PR #106, opened 2018-10-02
PA-2124 Updated curl version to 7.61.1

sebastian-miclea:
I've updated the version for curl, and tested locally on several OSes, but would need help on how to test all the combinations of runtime version vs platforms.
I know that I can add comments like "Jenkins please ...", but is this the correct way in testing all combinations?


MikaelSmith:
lgtm for client-tools

geoffnichols:
Hi @sebastian-miclea - currently, the best way to validate runtime changes is using a combination of the 'vanagon-generic-builder' and 'puppet-agent-extra' Ad Hoc jenkins pipelines. Please let me know if there's anything we can do to help there! :)

PR #108, opened 2018-10-04
(maint) Add patch for windows generated batch files to R2.5

bmjen:
This patch to the Ruby batch file generation is necessary when packaging
Ruby with Puppet's vanagon package builder, as used by Puppet agent,
Bolt, and PDK. When the msi installer is built, the path generated for
the installed Ruby cannot be assumed to be the same as the path in
the build environment. For that reason, we need this patch to use the
executables from the current working directory, so that it is always
using the vendored ruby installation.


bmjen:
Putting up a new PR for this. I tested the build of pdk-runtime with this branch, and results are here: https://builds.delivery.puppetlabs.net/pdk-runtime/b9572e0847fd98017887af462f0b06e6b23a8d91/artifacts/?C=M&O=D

I'm currently testing a build of PDK with that runtime.

bmjen:
This branch was able to successfully build pdk-runtime and apply the patch here:
https://builds.delivery.puppetlabs.net/pdk-runtime/b9572e0847fd98017887af462f0b06e6b23a8d91/artifacts/?C=M&O=D

And PDK was successfully built and tested using that runtime here:
https://builds.delivery.puppetlabs.net/pdk/97f141c35abb77c995a6710bc4cc67ea3b89d2a3/artifacts/?C=M&O=D

PR #109, opened 2018-10-08
(PDK-1056) Revert multi-version components to multiple single version components

rodjek:
Unfortunately, looping through the additional rubies in a single components doesn't result in the expected outcome due to the way the DSL writes out the component Makefiles. The `build` and `install` commands end up getting grouped together rather than getting executed in evaluation order.

So instead of:

1. build (2.1.9)
2. install (2.1.9)
3. build (2.5.1)
4. install (2.5.1)

We get:

1. build (2.1.9)
2. build (2.5.1)
3. install (2.1.9)
4. install (2.5.1)

In the case that @bmjen noticed on Friday, this lead to a version of `libselinux.so` that was linked against Ruby 2.5.1 being installed in the Ruby 2.1.9 runtime (as the 2.5.1 build command clobbered the artifacts of the 2.1.9 build command).

In order to unblock shipping PDK 1.7.1, this PR reverts the multi-ruby components back into single-ruby components. We can come back to this later and make it more efficient.


configs/projects/pdk-runtime.rb
```diff
@@ -186,11 +186,13 @@
...
       proj.component "ruby-#{rubyver}"
-    end
 
-    proj.component "ruby-multi-augeas" unless platform.is_windows?
-    proj.component "ruby-multi-selinux" if platform.name =~ /^el-(5|6|7)-.*/ || platform.is_fedora?
-    proj.component "ruby-multi-stomp"
+      ruby_minor = rubyver.split('.')[0,2].join('.')
+
+      proj.component "ruby-#{ruby_minor}-augeas" unless platform.is_windows?
+      proj.component "ruby-#{ruby_minor}-selinux" if platform.name =~ /^el-(5|6|7)-.*/ || platform.is_fedora?
```
bmjen: This needs to be updated to use `platform.is_el?` like the way it's done for the main ruby.

rodjek: Updated

PR #111, opened 2018-10-08
(maint) Fix curl 7.61.1 configure on AIX and Solaris 11

caseywilliams:
Changes in curl 7.61.1 produced errors while building all puppet-agent
runtimes on Redhat FIPS, Solaris 11, and AIX.

- Previously, Redhat FIPS would fall back to using the system OpenSSL
  when the Puppet's OpenSSL couldn't be found; this behavior seems to
  have disappeared, so set the Puppet OpenSSL path only for other
  platforms.
- Makefile genreation with automatic dependency tracking on Solaris 11
  and AIX now fails; disable it, since these are one-time builds.


caseywilliams:
(updated to fix a commit message typo)

PR #112, opened 2018-10-10
(BOLT-380) Update net-ssh to 5.0.2 for bolt-runtime

MikaelSmith:
Upgrading primarily to enable ed25519 support in packaging.


MikaelSmith:
This will require synchronization with some changes in bolt and bolt packaging. So hold off until we're ready to do all of them.

PR #114, opened 2018-11-01
(PA-2279) Patch Ruby CVE 2018-16395

caseywilliams:
Patches Ruby 2.1.9, 2.4.4, and 2.5.1 for CVE-2018-16395 (corrects name equality checks for `OpenSSL::X509::Name`) -- note that puppet does not appear to use this codepath, but this is worth fixing.

Some links:
- [The CVE](https://www.ruby-lang.org/en/news/2018/10/17/openssl-x509-name-equality-check-does-not-work-correctly-cve-2018-16395/)
- [The fix for it in Ruby](https://github.com/ruby/ruby/commit/93bc10272734cbbb9197470ca629cc4ea019f6f0)


caseywilliams:
We'll just update to ruby 2.5.3 instead -- I'll open a new PR for that.

PR #118, opened 2018-12-06
(PA-2348) Remove Fedora 27 x86_64 from the puppet-runtime repository.

tavilarion:
Removed Fedora 27 x86_64


bmjen:
@tavilarion  @geoffnichols This isn't part of the originating ticket, but if we are removing 27 from this repo entirely, should Fedora 25 and 26 be removed as well?

tavilarion:
@bmjen @geoffnichols  I also thought about it, but since I am a new member of the team and currently not fully aware of all implications, I decided to stick with the ticket. Would be happy to remove all the older versions if it is decided to do so.

tavilarion:
@speedofdark this is done.

ekinanp:
We had to revert this change with https://github.com/puppetlabs/puppet-runtime/pull/127. For future reference, a platform should only be removed from puppet-runtime if we know for sure that it is no longer supported in everything that depends on puppet-runtime. Since Fedora 27 is still built with the 1.10.x and 5.3.x series, we still need it in puppet-runtime.

PR #122, opened 2018-12-13
added osx-10.14

loopinu:


configs/projects/agent-runtime-5.5.x.rb
```diff
@@ -39,6 +39,6 @@
   proj.component 'rubygem-trollop'
   proj.component 'rubygem-hiera-eyaml'
   proj.component 'ruby-stomp'
-  proj.component 'yaml-cpp' if platform.name =~ /osx-10.12|fedora-29/
+  proj.component 'yaml-cpp' if platform.name =~ /osx-10.12|osx-10.14|fedora-29/
```
ekinanp: Interesting, this does not include OSX 10.13. @caseywilliams or @McdonaldSeanp why is that?

ekinanp:
Can we change the commit message to `(PA-2219) Add Mac OS 10.14 Mojave`

configs/platforms/osx-10.14-x86_64.rb
```diff
@@ -0,0 +1,29 @@
+platform 'osx-10.14-x86_64' do |plat|
+    plat.servicetype 'launchd'
+    plat.servicedir '/Library/LaunchDaemons'
+    plat.codename 'sierra'
```
ScottGarman: Looks like the codename should be updated to 'mojave'

PR #123, opened 2018-12-13
(PDK-1050) Updates to support Fedora 28 and 29 for pdk-runtime

scotje:


scotje:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-runtime/view/pdk-runtime/job/platform_pdk-runtime_runtime-vanagon-packaging_master/73/ is in progress

configs/components/runtime-pdk.rb
```diff
@@ -33,7 +33,15 @@
...
   else # Linux and Solaris systems
-    libdir = "/opt/pl-build-tools/lib64"
+    if platform.name =~ /fedora-29/
+      # On newer linux platforms we are no longer using the pl-build-tools
+      # package and instead relying on standard distribution versions of
+      # gcc, etc.
+      libdir = "/usr/lib64"
+    else
+      libdir = "/opt/pl-build-tools/lib64"
+    end
```
scotje: I think ideally `libdir` would come from platform config directly. @caseywilliams is there anything currently stored in the platform config we could use?

caseywilliams: All the information we get about a platform beyond the very basics comes directly from this repo. Vanagon's job has not really been to maintain any information about where things are (beyond what's absolutely required to run make and build/package software) -- most of its platform capabilities are focused on the distro type, rather than any specific release of a distro -- [there are some examples here](https://github.com/puppetlabs/vanagon/blob/5232dd1ea33513fb3196db3ab3b285e381b416de/lib/vanagon/platform.rb#L342).

So, setting the libdir as shown here is the standard way to do it -- perhaps storing the default libdir in vanagon itself would make sense, but I'm not aware of any existing efforts to do that.

scotje: OK, so there's no equivalent to something like this:

```
  plat.servicedir "/usr/lib/systemd/system"
  plat.defaultdir "/etc/sysconfig"
```

in the platform config files for a `libdir` type variable?

That's fine for now although it seems like it could be useful as we migrate more things away from pl-build-tools and the `libdir` has the potential to vary more from distro to distro.

scotje:
@bmjen @caseywilliams the above-linked build passed all platforms (including newly added fedora-28 and fedora-29)

PR #128, opened 2019-01-10
(PA-2334) do not use root for homebrew on OSX

mihaibuzgau:
This PR also removes osx-10.10 and osx-10.11 as they are not longer support platforms for versions of Puppet >= 5.5

PR #131, opened 2019-01-25
Revert "(PA-2334) do not use root for homebrew on OSX"

caseywilliams:
Reverts puppetlabs/puppet-runtime#128

CI for puppet-agent 1.10.x depends on macOS 10.10 and 10.11, and we haven't committed to removing it there. We'll need to keep these platforms in this repo until the puppet-agent 1.10.x branch is retired.


geoffnichols:
Merging per discussion with @caseywilliams and @McdonaldSeanp

PR #132, opened 2019-01-25
Revert "Merge pull request #131 from puppetlabs/revert-128-master"

caseywilliams:
This reverts commit 1b752d605dc8021c931fc956997a136d6e338db7, reversing
changes made to 933c99524d929361b5a88b86de19a9146de1042d.

The changes to the projects and platforms here are fine, but the removal
of OS X 10.10 and 10.11 should happen separately or not at all.

I'm going to run a few tests on all the macOS platforms to make sure everything's still okay before I remove the do not merge label.


caseywilliams:
I've built and tested this for all the macOS platforms supported in in agent 1.10.x and 5.5.x -- I think it's ready to go!

PR #133, opened 2019-01-29
(BOLT-978) Dont use pl-build-tools for bolt-runtime on fedora 29

donoghuc:
As part of the work packaging and shipping puppet-bolt for fedora we use build tools that exist in the newer distros.

### DIFF
--- a/configs/components/runtime-bolt.rb
+++ b/configs/components/runtime-bolt.rb
@@ -13,9 +13,9 @@ component "runtime-bolt" do |pkg, settings, platform|
     pkg.install_file "#{settings[:tools_root]}/bin/libgdbm_compat-4.dll", "#{settings[:ruby_bindir]}/libgdbm_compat-4.dll"
     pkg.install_file "#{settings[:tools_root]}/bin/libiconv-2.dll", "#{settings[:ruby_bindir]}/libiconv-2.dll"
     pkg.install_file "#{settings[:tools_root]}/bin/libffi-6.dll", "#{settings[:ruby_bindir]}/libffi-6.dll"
-  elsif platform.is_macos? or platform.name =~ /sles-15/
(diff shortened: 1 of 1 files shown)


donoghuc:
Exploring adding fedora 29 and 28. This change is inspired by https://github.com/puppetlabs/puppet-runtime/pull/123/files/9f83e4e96baa8bc0612bb0bddb5b5345f4d0e4fc#diff-97f3c8a8cc5d403c4b864a0148cbdcec

Testing with https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-runtime/view/bolt-runtime/job/platform_bolt-runtime_runtime-vanagon-init_master/67/

configs/components/runtime-bolt.rb
```diff
@@ -19,7 +19,7 @@
 
   else # Linux and Solaris systems
     libbase = platform.architecture =~ /64/ ? 'lib64' : 'lib'
-    libdir = "/opt/pl-build-tools/#{libbase}"
+    libdir = platform.name =~ /fedora-29/ ? "/usr/lib64" : "/opt/pl-build-tools/#{libbase}"
```
caseywilliams: Oh gosh, actually, I just remembered -- would this work if you added fedora-29 to the conditional just above (https://github.com/puppetlabs/puppet-runtime/pull/133/files#diff-8da58944b3f541d50ae6f6257ad40b8cL16) instead? SLES 15 is another platform that doesn't use pl-build-tools, since its compiler is new enough (also, if that does work, adding a more descriptive comment than is currently there would surely help)

donoghuc: Thanks for the review. I will try an ad-hoc build with your suggestion. I agree that is less than helpful.

PR #136, opened 2019-02-04
(maint) Bump minitar version for bolt runtime

donoghuc:
Bolt shares a dependency with puppet-runtime on the minitar gem. In order to build pe-bolt-server we were also adding the newer version in the bolt-vanagon project. This resulted in puppet-bolt and pe-bolt-server packages having different minitar versions. This commit allows us to specify the latest minitar version in bolt-runtime without affecting other projects that may use the older gem.

### DIFF
--- a/configs/components/rubygem-minitar.rb
+++ b/configs/components/rubygem-minitar.rb
@@ -1,8 +1,19 @@
 component "rubygem-minitar" do |pkg, settings, platform|
+  # Projects may define a :rubygem_minitar_version setting, or we use 0.6.1 by default:
+  version = settings[:rubygem_minitar_version] || '0.6.1'
+  pkg.version version
+
+  case version
+  when "0.6.1"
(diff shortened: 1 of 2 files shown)


donoghuc:
Corresponding bolt-vanagon change: https://github.com/puppetlabs/bolt-vanagon/pull/74

Also still testing locally.

configs/components/rubygem-minitar.rb
```diff
@@ -1,8 +1,19 @@
 component "rubygem-minitar" do |pkg, settings, platform|
+  # Projects may define a :rubygem_minitar_version setting, or we use 0.6.0 by default:
```
MikaelSmith: Comment should say `0.6.1 by default`.

donoghuc:
Updated comment with correct version and tested building both puppet-bolt and pe-bolt-server.

PR #143, opened 2019-03-07
(PA-2491)-Remove SLES11/12 s390x from the puppet-runtime repository

loopinu:
Removed Sles 11/12 s390x refrences from the files and also any logic that included s390x because there is no need for them anymore as rhel s390x is also removed.

NOTE: has dependency with PR: https://github.com/puppetlabs/puppet-runtime/pull/140


geoffnichols:
It looks like the related agent changes have merged up to master - is this good to merge now?

caseywilliams:
Yes, ready!

PR #145, opened 2019-03-14
(BOLT-495) Update net-ssh to 5.2.0

donoghuc:
Upgrading to pull in updates that will improve host-key-check option for bolt.


donoghuc:
Ready for review, but please let me merge it. Thanks

PR #149, opened 2019-03-28
PE-26080 Update directory path for el-8 in pe-installer runtime

shaigy:
Update libdir path to /usr/lib64 on el-8 in pe-installer runtime

### DIFF
--- a/configs/components/runtime-pe-installer.rb
+++ b/configs/components/runtime-pe-installer.rb
@@ -19,7 +19,11 @@ component "runtime-pe-installer" do |pkg, settings, platform|
 
   else # Linux and Solaris systems
     libbase = platform.architecture =~ /64/ ? 'lib64' : 'lib'
-    libdir = "/opt/pl-build-tools/#{libbase}"
+    if platform.name =~ /el-8/
+      libdir = "/usr/#{libbase}"
+    else
+      libdir = "/opt/pl-build-tools/#{libbase}"
+    end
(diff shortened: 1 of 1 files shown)


shaigy:
Built pe-installer runtime locally for el-8 successfully with the change.

configs/components/runtime-pe-installer.rb
```diff
@@ -19,7 +19,11 @@
 
   else # Linux and Solaris systems
     libbase = platform.architecture =~ /64/ ? 'lib64' : 'lib'
-    libdir = "/opt/pl-build-tools/#{libbase}"
+    if platform.name =~ /el-8/
+      libdir = "/usr/#{libbase}"
+    else
+      libdir = "/opt/pl-build-tools/#{libbase}"
```
highb: I don't think we need to continue using `pl-build-tools` for any platform that we build.

highb: @McdonaldSeanp @caseywilliams Should we remove all use of `pl-build-tools` from the runtimes?

shaigy: pl-build-tools are still used here https://github.com/puppetlabs/puppet-runtime/blob/8736f3c83e8e39a683cacfe8195f9e5b9d87c12d/configs/components/runtime-pdk.rb#L35-L43

highb: Alright... I guess we'll keep using it? Hopefully it doesn't end up breaking suddenly later.

PR #155, opened 2019-04-05
(PA-2404)-Add debian 10 buster support

loopinu:


configs/platforms/debian-10-amd64.rb
```diff
@@ -0,0 +1,23 @@
+platform "debian-10-amd64" do |plat|
+  plat.servicedir "/lib/systemd/system"
+  plat.defaultdir "/etc/default"
+  plat.servicetype "systemd"
+  plat.codename "buster"
+
+  plat.provision_with "export DEBIAN_FRONTEND=noninteractive; apt-get update -qq; apt-get install -qy --no-install-recommends build-essential devscripts make quilt pkg-config debhelper rsync fakeroot cmake"
```
denis-susan: why not add these packages to the packages array and just run one provision_with command?

PR #156, opened 2019-04-05
(maint) Updates for el-8 for client-tools

shrug:


configs/components/runtime-client-tools.rb
```diff
@@ -16,8 +16,10 @@
     # Do nothing
 
   else # Linux and Solaris systems
-    libbase = platform.architecture =~ /64/ ? 'lib64' : 'lib'
-    libdir = "/opt/pl-build-tools/#{libbase}"
+    if platform.name !~ /el-8-x86_64/
+      libbase = platform.architecture =~ /64/ ? 'lib64' : 'lib'
+      libdir = "/opt/pl-build-tools/#{libbase}"
+    end
```
mwaggett: What happens here for el 8? Looks like `libdir` will be unset, but is needed below in the `install` step.

shrug: it looks like we're overriding it here, it's originally set in https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/_shared-client-tools-runtime.rb#L43

mwaggett: Wouldn't that have to be referred to as `settings[:libdir]` in this file? I'm not understanding how that override happens..

PR #157, opened 2019-04-08
(maint) Add el-8-x86_64 to build_defaults

shrug:


mwaggett:
@shrug I think it needs to be added to `rpm_targets` too?

shrug:
yep! good catch thank you

PR #160, opened 2019-04-15
(maint) Add Dockerfile and Makefile for building runtime containers

underscorgan:


Iristyle:
Hmm... there are a bunch of errors in the Travis log, but it still went green. I wonder if there's a minor issue with how we're doing error detection there.

underscorgan:
@Iristyle yep, travis needs some updating https://github.com/puppetlabs/puppet-runtime/pull/185

joshcooper:
I know little about the Dockerfile, but general 👍 from me.

@mihaibuzgau heads up, we want to publish `puppet-runtime` docker images for a few different reasons. One is to simplify running puppet (and friends) acceptance tests, eg. run beaker tests locally against a running `puppet-runtime` docker container.

PR #162, opened 2019-04-18
(maint) Remove win32-eventlog from modern agents

donoghuc:
According to PA-1129 / PUP-5756 we dont need this gem anymore. This commit removes it from shared agent components and only adds it to 1.10.x runtime.

### DIFF
--- a/configs/projects/agent-runtime-1.10.x.rb
+++ b/configs/projects/agent-runtime-1.10.x.rb
@@ -57,4 +57,5 @@ project 'agent-runtime-1.10.x' do |proj|
 
   proj.component 'rubygem-gettext-setup'
   proj.component 'ruby-stomp'
+  proj.component 'rubygem-win32-eventlog' if platform.is_windows?
 end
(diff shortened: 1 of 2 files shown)


configs/projects/agent-runtime-1.10.x.rb
```diff
@@ -57,4 +57,5 @@
 
   proj.component 'rubygem-gettext-setup'
   proj.component 'ruby-stomp'
+  proj.component 'rubygem-win32-eventlog' if platform.is_windows?
```
donoghuc: Is this necessary?

donoghuc:
@sebastian-miclea @gimmyxd While evaluating what components to ship with bolt runtime I discovered that the `win32-eventlog` gem does not seem to be needed (at least in modern agents, I don't have a sense of what the 1.10.x runtime needs). https://tickets.puppetlabs.com/browse/PUP-5756 https://tickets.puppetlabs.com/browse/PA-1129

configs/projects/agent-runtime-1.10.x.rb
gimmyxd: I'm not really familiar with what is still needed on 1.10.x, but since it won't be around for that much we can let this here to not brake anything.

PR #171, opened 2019-05-08
(ORCH-2296) Include net-ssh optional dependencies for *nix

donoghuc:
The PR to transition the gems included in bolt-vanagon to bolt-runtime missed optionally including these gems on non-windows platforms.


donoghuc:
Missed these from transition. https://github.com/puppetlabs/bolt-vanagon/blob/d230d094275ba7b7ff862483921d572accd7c2bc/configs/projects/bolt-shared.rb#L6-L11

donoghuc:
I filed https://tickets.puppetlabs.com/browse/BOLT-1296 with some specifics on what is blocking including these gems on windows.

I am going to go ahead and merge this because we need to continue shipping these with *nix.

PR #174, opened 2019-05-13
(PA-2191) winruby UTF8 Fallback for no CodePage

jcoconnor:
This is to deal specifically with setting the Region to Arabic which
sets the codepage to 720. This codepage isn't recognises by Ruby which
causes it to fail during initialisation when setting up LOCALE.

So patch Ruby for Windows Only to Fallback to UTF for THIS ERROR ONLY.
This also applies to any other languages that use a codepage not
recognised by Ruby.

### DIFF
--- a/configs/components/ruby-2.4.5.rb
+++ b/configs/components/ruby-2.4.5.rb
@@ -68,6 +68,7 @@ component 'ruby-2.4.5' do |pkg, settings, platform|
     pkg.apply_patch "#{base}/update_rbinstall_for_windows.patch"
     pkg.apply_patch "#{base}/PA-1124_add_nano_server_com_support-8feb9779182bd4285f3881029fe850dac188c1ac.patch"
     pkg.apply_patch "#{base}/windows_socket_compat_error_r2.4.patch"
+    pkg.apply_patch "#{base}/windows_nocodepage_utf8_fallback_r2.4.patch"
   end
 
   ####################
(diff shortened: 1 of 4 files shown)


gimmyxd:
puppet-agent 5 uses ruby 2.4.5. I think we should also patch that since the bug was reported on 5.5.3

jcoconnor:
@gimmyxd I've added for 2.4.5 also.

caseywilliams:
This looks looks right to me - I've kicked off test builds in Jenkins using the [generic vanagon pipeline](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/) for agent runtimes for [5.5.x](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/413/), [6.0.x](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/414/), [6.4.x](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/415/), and [master](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/416/). When those finish successfully, they'll be shipped to the build server and available for testing with puppet-agent's branches.

jcoconnor:
Have verified that the Ruby ```registry.rb``` file is updated correctly and the ```puppet``` command executes on an Arabic Region Windows 2016 Server using ````Puppet Agent 5.5```

Just waiting for test suite to complete.

jcoconnor:
This is ready for merge - passes testing on Puppet-Agent 5.5 and works on Arabic systems with Codepage set to 720

joshcooper:
If it hasn't been done already, could someone file a bug with ruby (https://bugs.ruby-lang.org) and submit the patch upstream?

jcoconnor:
Good point will log a ticket for sprint planning for this - see [PA-2690](https://tickets.puppetlabs.com/browse/PA-2690)

PR #180, opened 2019-05-30
(PDK-1378) Configure pdk-runtime git to use /etc/gitconfig

scotje:


scotje:
@underscorgan This shouldn't have any impact on non-`pdk-runtime` builds but wanted someone outside of our immediate team to be aware of the change.

PR #184, opened 2019-06-05
(ct-66) Update curl to 7.65.1 to address CVEs

speedofdark:
See https://curl.haxx.se/changes.html for a full list of addressed
items in 7.65.1.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -1,6 +1,6 @@
 component 'curl' do |pkg, settings, platform|
-  pkg.version '7.64.0'
-  pkg.md5sum 'a026740d599a32bcbbe6e70679397899'
+  pkg.version '7.65.1'
+  pkg.md5sum '6f31ec164741cc6aaac461d4b38ffdbe'
   pkg.url "https://curl.haxx.se/download/curl-#{pkg.get_version}.tar.gz"
   pkg.mirror "#{settings[:buildsources_url]}/curl-#{pkg.get_version}.tar.gz"
 


configs/components/curl.rb
```diff
@@ -1,6 +1,6 @@
 component 'curl' do |pkg, settings, platform|
-  pkg.version '7.64.0'
-  pkg.md5sum 'a026740d599a32bcbbe6e70679397899'
+  pkg.version '7.65.0'
```
mwaggett: no idea if y'all care, but 7.65.1 came out today if you want to bump to that while you're at it. ¯\_(ツ)_/¯

caseywilliams: Yes, this is worth doing

geoffnichols: +1

geoffnichols:
@speedofdark would you include a link to a passing build with this change? I believe a passing run in the [platform "ad-hoc" pipeline](https://confluence.puppetlabs.com/display/AGENT/How+to+run+the+ad-hoc+puppet-agent+pipeline) would be ideal, but a passing [jenkins PR test](https://confluence.puppetlabs.com/display/AGENT/How+to+run+PR+tests+on+puppet-agent+and+its+components) may be the minimum level of validation we'd want to see here. 😄

configs/components/curl.rb
speedofdark: Updated to 7.65.1.

speedofdark:
Ran against "el-7-x86_64 ubuntu-16.04-amd64 aix-6.1-ppc ubuntu-16.04-ppc64el windows-2012r2-x64 solaris-11-i386" and came out green: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/448/

PR #186, opened 2019-06-14
(maint) Add redhatfips-7-x86_64 to build_defaults

sarameisburger:


caseywilliams:
:thinking: We already get redhatfips tarballs shipped for puppet-agent, even though this hasn't been here -- what does this do when the repo only builds tarballs?

PR #187, opened 2019-06-17
(MAINT) Remove PDK restriction on latest puppet for Ruby 2.5

scotje:
There is currently no reason to cap the puppet version for Ruby 2.5.x, we might as well let packaging resolve it to the latest. This setting is only used in the downstream `pdk-vanagon` repo.

### DIFF
--- a/configs/projects/pdk-runtime.rb
+++ b/configs/projects/pdk-runtime.rb
@@ -50,6 +50,10 @@ project 'pdk-runtime' do |proj|
   proj.setting(:ruby_version, "2.4.5")
   proj.setting(:ruby_api, "2.4.0")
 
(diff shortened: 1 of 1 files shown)


scotje:
Step one of resolving Puppet gem version mismatch in `pdk-vanagon` build.

scotje:
Updated to set a `latest_puppet` for the default Ruby 2.4.5 runtime as well.

scotje:
Successful build: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-runtime/view/pdk-runtime/job/platform_pdk-runtime_runtime-vanagon-packaging_master/91/

PR #191, opened 2019-07-02
(WIP)(BOLT-1339) add redhatfips support for PE bolt server

speedofdark:
This commit makes using openssl vs. system supplied libraries conditional
based on the platform being FIPS or not. FIPS uses system libraries.

### DIFF
--- a/configs/projects/bolt-runtime.rb
+++ b/configs/projects/bolt-runtime.rb
@@ -103,11 +103,20 @@ project 'bolt-runtime' do |proj|
     '-DOPENSSL_NO_HEARTBEATS',
   ])
 
+  if platform.name =~ /^redhatfips-7-.*/
+    # Link against the system openssl instead of our vendored version:
+    proj.setting(:system_openssl, true)
+  end
+
   # What to build?
   # --------------
 
   # Ruby and deps
-  proj.component "openssl-#{proj.openssl_version}"
+  # BOLT-1339 use system SSL for FIPS
+  unless proj.settings[:system_openssl]
+    proj.component "openssl-#{proj.openssl_version}"
+  end
+  
   proj.component "runtime-bolt"
   proj.component "puppet-ca-bundle"
   proj.component "ruby-#{proj.ruby_version}"


configs/projects/bolt-runtime.rb
```diff
@@ -107,7 +107,7 @@
   # --------------
 
   # Ruby and deps
-  proj.component "openssl-#{proj.openssl_version}"
+  proj.component "openssl-#{proj.openssl_version}" unless platform.name =~ /^redhatfips-7-.*/  # BOLT-1339 FIPS uses system SSL libs
```
caseywilliams: I think you'll need to do a bit more than this to make sure curl, etc. build correctly. Try grepping this repo for `:system_openssl` -- the agent runtimes use that setting to do this filtering instead of a specific platform name.

speedofdark: That setting just controls a conditional at a lower level that includes the component or not.

configs/platforms/redhatfips-7-x86_64.rb
```diff
@@ -14,7 +14,6 @@
     "libsepol-devel",
     "libselinux-devel",
     "make",
-    "openssl-devel",
```
caseywilliams: this is a system package, so we do want it -- this is what's replacing our own custom-built ssl

configs/projects/bolt-runtime.rb
caseywilliams: You will still need to update the ca bundle component to use the system openssl. It already look for the `:system_openssl` setting -- did you try the grep?

speedofdark: In progress.

geoffnichols:
Just to confirm, are there any changes needed to any of the pe-bolt-server project configs?

speedofdark:
Yes. WIP.

donoghuc:
I don't think you need to modify bolt-runtime for pe-bolt-server to use system ssl because bolt is not installed in PE. A change like this will be necessary for building bolt for fips.

speedofdark:
I'm thinking that as well. pe-bolt-server does not appear to use openssl at all.

PR #195, opened 2019-07-11
(PDK-1397) Build pdk-runtime on fedora 30

rodjek:
Changes this conditional so that we don't have to change it again for future fedora or rhel releases

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-runtime/view/pdk-runtime/job/platform_pdk-runtime_runtime-vanagon-packaging_master/BUILD_TARGET=fedora-30-x86_64,SLAVE_LABEL=worker/95/console

### DIFF
--- a/configs/components/runtime-pdk.rb
+++ b/configs/components/runtime-pdk.rb
@@ -33,7 +33,8 @@ component "runtime-pdk" do |pkg, settings, platform|
     # Do nothing
 
   else # Linux and Solaris systems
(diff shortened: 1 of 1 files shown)


glennsarti:
@rodjek Not sure if you need any other reviewers given that runtime affects many downstream projects.

scotje:
> @rodjek Not sure if you need any other reviewers given that runtime affects many downstream projects.

The particular file changed here should only impact the pdk-runtime artifact.

PR #196, opened 2019-07-11
(maint) Update net-ssh dependencies for pe-installer-runtime

barriserloth:
This commit updates the dependencies for pe-installer-runtime, so that
net-ssh is able to properly handle the new default format for private
keys shipped in the version of OpenSSH that comes as the default on el8
platforms. This is (again, unfortunately) for our version of Bolt, which
isn't installed as a package and as such does not come with some of the
handy dependencies that should be present.

### DIFF
--- a/configs/projects/pe-installer-runtime.rb
+++ b/configs/projects/pe-installer-runtime.rb
@@ -128,6 +128,14 @@ project 'pe-installer-runtime' do |proj|
   proj.setting(:rubygem_net_ssh_version, '5.2.0')
   proj.component 'rubygem-net-ssh'
 
+  # net-ssh dependencies for el8's OpenSSH default key format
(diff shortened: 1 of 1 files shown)


barriserloth:
I have locally built the new pe-installer-runtime package using the following command: `be build pe-installer-runtime el-8-x86_64` and verified that it succeeds and contains the newly added gems in the output (a snippet of the installed gems json is below):

```
"rubygem-bcrypt_pbkdf": {\n "version": "1.0.0",\n "url": "https://rubygems.org/downloads/bcrypt_pbkdf-1.0.0.gem"\n },\n "rubygem-ed25519": {\n "version": "1.2.4",\n "url": "https://rubygems.org/downloads/ed25519-1.2.4.gem"\n }
```

PR #197, opened 2019-07-15
(maint) Add aws-sdk-ec2 gem to bolt projects

lucywyman:
We now depend on the aws-sdk-ec2 gem, as of
https://github.com/puppetlabs/bolt/pull/1073, which requires adding it
and it's dependencies to the bolt-runtime and pe-bolt-server-runtime


caseywilliams:
Thanks @lucywyman, this looks right to me!

PR #204, opened 2019-08-07
(maint) Added openjdk 11 for ubuntu 18.04

loopinu:
This adds open jdk 11 that is needed to use java ca ceterificates using keytool


Iristyle:
Hey @loopinu / @GabrielNagy I think we should look at a different way of accomplishing this goal and revert this PR for now. The introduction of this package has broken puppetserver - see https://github.com/puppetlabs/pe-puppetserver/pull/364 and https://tickets.puppetlabs.com/browse/PE-27036 for more information

GabrielNagy:
Hi,

https://github.com/puppetlabs/puppet-runtime/commit/f5187aeac552d96e22ae1fd706c80c153eee53bb should fix the issue, we managed to reproduce and fix the issue by changing java11 to java8.

PR #211, opened 2019-08-21
(MAINT) Fix docker test

nwolfe:
We need to pull the exit code off the status object before testing.

Also silence the git fetch warning when you're up to date,
and remove the temporary testing container so we don't leave a bunch of
dead containers lying around afterwards.


docker/puppet-runtime/spec/container_spec.rb
```diff
@@ -1,14 +1,14 @@
...
-  include Pupperware::SpecHelpers
-
   before(:all) do
     @image = require_test_image
   end
 
   it 'should have ruby' do
     result = run_command("docker run #{@image} /opt/puppetlabs/puppet/bin/ruby --version")
-    expect(result[:status]).to eq(0)
+    expect(result[:status].exitstatus).to equal(0)
```
Iristyle: Interesting, I thought `equal` was for identity comparison? But it works for integers of the same value?

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/equality-matchers#compare-using-equal-(equal?)
https://apidock.com/ruby/Object/equal%3F

docker/.rspec
```diff
@@ -1,4 +1,4 @@
---require pupperware/spec_helper
 --format RspecJunitFormatter
 --out TEST-rspec.xml
 --format documentation
+--fail-fast
```
Iristyle: I'm not sure we want `--fail-fast` here? There were a few suites (pe-client-tools-vanagon, puppet-agent, puppet-runtime and r10k) where I didn't add it in my recent sweep, because the suite passing isn't contingent on a few preflight things working first.

Since we only have one test in the suite it also makes no practical difference at the moment.

nwolfe: Ok I'll take it off, I don't have an opinion on using it, I just remember you said "we should use --fail-fast everywhere" a few days ago.

Iristyle: Yeah, I think my main thinking was catching completely fatal errors in the initial setup and killing the suite as soon as we know. So instead of letting a suite run for 10 minutes in a degraded / unusable state, just let it die.

I wonder if we could also use Ruby `fatal` to do the same thing?

nwolfe:
@Iristyle updated; reverted the comparison method so the only test change here is adding the `.exitstatus` call.

Also:
- silence the git fetch warning when you're up to date
- add `--rm` flag to clean up the `docker run ...` containers left behind by the test
- rename the file from `dockerfile_spec.rb` to `container_spec.rb`

PR #212, opened 2019-08-23
(PA-2856) Remove Ubuntu 14.04 from puppet-runtime

Dorin-Pleava:
This PR removes Ubuntu 14.04 platform from puppet-runtime together with
some specific casing.


gimmyxd:
Usually we start dropping support for a platform after it reaches end of life.

@donoghuc here is the epic for ubuntu removal: https://tickets.puppetlabs.com/browse/PC-1191
It was already deprecated from PE and we are planning to also remove agent FOSS support in the next days. Since this is also affecting bolt/pdk we can just remove ubuntu 14.04 from agent pipeline and merge this when all other projects are ready.

PR #218, opened 2019-09-04
(PA-2838) add windowsfips-2012r2-x64

ciprianbadescu:


caseywilliams:
LGTM!

McdonaldSeanp:
A note about windows FIPS: I don't know if @puppetlabs/night-s-watch has already looked in to it, but @shrug and I have discussed what's required for FIPS on windows, and never came to a good conclusion about whether or not FIPS for windows required use of Microsoft's MSVC compiler when building OpenSSL to be compliant.

If we someone else already determined that it's okay to use MinGW then go ahead and ignore me, but I just wanted to bring it up because I had not heard anything about folks finding an explicit answer about it.

ciprianbadescu:
Hi,

If we switch openssl-mips to MSVC compiler we will need to move to MSVC all dependencies as I understood that there is no binary compatibility between MSVC and gcc(mingw): openssl/curl/ruby.
I am not sure about facter.

Bellow my understanding of https://www.openssl.org/docs/fips/UserGuide-2.0.pdf, but confirmation is needed as this is sensitive matter:
> Chapter 3:
> Any computing platform that meets the conditions in the Security Policy can be used to host a FIPS
> 140-2 validated FIPS Object Module provided that module is generated in accordance with the Security Policy.
> 
> At the time the OpenSSL FIPS Object Module v2.0 was developed, all Unix-like environments 
> supported by the full OpenSSL distribution were also supported by the FIPS validated source files 
> included in the FIPS Object Module.

The question marks I have are:
 - does the change applied to remove '\r' and needed during openssl compilation breaks openssl fips compatibility? - my assumption is that, as the change is not done for fips module compilation itself, should not interfere with fips module validation 
 - how do we implement "4.1.3 Verifying Integrity of the Full Distribution for the FIPS Object Module"

PR #219, opened 2019-09-04
Ship updated gem components for bolt/bolt-server runtimes

donoghuc:
Keep up to date with latest gem releases for bolt/bolt-server packages.

### DIFF
--- a/configs/components/rubygem-addressable.rb
+++ b/configs/components/rubygem-addressable.rb
@@ -1,6 +1,6 @@
 component 'rubygem-addressable' do |pkg, _settings, _platform|
-  pkg.version '2.6.0'
-  pkg.md5sum '33055ce32b4434fab232b57d0a0fd360'
+  pkg.version '2.7.0'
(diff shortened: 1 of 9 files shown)


donoghuc:
I notice that `gettext-setup` has an update too, (`0.29 -> 0.30') https://github.com/puppetlabs/gettext-setup-gem/compare/0.30...0.31 . But that is used by more than just bolt. Is that something we want to pull in across the board?

caseywilliams:
oh, I didn't see your gettext command - I'd want to just go ahead and update it globally, but if people are nervous about that we could introduce the versioning pattern we're using for some other components for gettext

lucywyman:
I agree with updating globally - it seems easy enough to revert and rebuild if something goes wrong.

donoghuc:
Ok, I will follow up with @puppetlabs/night-s-watch about updating `gettext-setup`. For now, the changes to that gem don't impact bolt at all so i'm going to merge this so we can get packages built for release.

PR #220, opened 2019-09-11
(PUP-10013) Update minitar to 0.9

jtappa:
We can use the new options hash in minitar 0.9 to set `fsync=false`which should speed things up for folks

### DIFF
--- a/configs/components/rubygem-minitar.rb
+++ b/configs/components/rubygem-minitar.rb
@@ -1,16 +1,6 @@
-component "rubygem-minitar" do |pkg, settings, platform|
-  # Projects may define a :rubygem_minitar_version setting, or we use 0.6.1 by default:
-  version = settings[:rubygem_minitar_version] || '0.6.1'
-  pkg.version version
-
-  case version
-  when "0.6.1"
-    pkg.md5sum "ce4ee63a94e80fb4e3e66b54b995beaa"
-  when "0.8"
-    pkg.md5sum "af220249c7dfe1774de3a81da4bb21d7"
-  else
-    raise "rubygem-minitar version #{version} has not been configured; Cannot continue."
-  end
+component 'rubygem-minitar' do |pkg, settings, platform|
+  pkg.version '0.9'
+  pkg.md5sum '4ab2c278183c9a83f3ad97066c381d84'
 
   instance_eval File.read('configs/components/_base-rubygem.rb')
 end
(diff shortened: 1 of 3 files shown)


configs/components/rubygem-minitar.rb
```diff
@@ -1,11 +1,11 @@
 component "rubygem-minitar" do |pkg, settings, platform|
   # Projects may define a :rubygem_minitar_version setting, or we use 0.6.1 by default:
-  version = settings[:rubygem_minitar_version] || '0.6.1'
+  version = settings[:rubygem_minitar_version] || '0.9'
   pkg.version version
 
   case version
-  when "0.6.1"
-    pkg.md5sum "ce4ee63a94e80fb4e3e66b54b995beaa"
+  when "0.9"
```
donoghuc: Bolt and bolt server added the version logic to get the newest minitar. If shared_agent component is on latest we can just get rid of the logic and have a single version. That would involve just deleting the lines in https://github.com/puppetlabs/puppet-runtime/blob/0c65b6f1cc6e178ed607f900dfde67402f4c7a65/configs/projects/bolt-runtime.rb#L11 and https://github.com/puppetlabs/puppet-runtime/blob/0c65b6f1cc6e178ed607f900dfde67402f4c7a65/configs/projects/_shared-pe-bolt-server.rb#L45 and you could change this to something like
```
component 'rubygem-minitar' do |pkg, settings, platform|
  pkg.version '0.9'
  pkg.md5sum '4ab2c278183c9a83f3ad97066c381d84'

  instance_eval File.read('configs/components/_base-rubygem.rb')
end
```

joshcooper:
@mihaibuzgau @gimmyxd does "jenkins please ..." work on runtime PRs?

It would be good to get a build of puppet-runtime with this change and make sure the gem is in same directory as it was before (as there are two gem directories, the normal one and the one shared by puppetserver). On Windows puppet agent 6.9.0 shows:

```
Administrator@b7eoxu6c4ijfz1k /cygdrive/c/Program Files/Puppet Labs
$ find Puppet/ -type d -name "minitar*"
Puppet/puppet/lib/ruby/gems/2.5.0/gems/minitar-0.6.1
Puppet/puppet/lib/ruby/gems/2.5.0/gems/minitar-0.6.1/lib/archive/tar/minitar
```

gimmyxd:
@joshcooper @jtappa should work(i don't think i used it), but we usually use vanagon-generic-builder for cases like this: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-generic-builder-init_generic-builder/

gimmyxd:
jenkins please test this,puppet-agent

joshcooper:
Per @mihaibuzgau, this is ok to merge now

PR #227, opened 2019-09-23
(maint) fix net_http_eof patch apply

gimmyxd:
Fix net_http_eof patch apply for multiple rubies.
This should fix the pdk-runtime build.


gimmyxd:
@scotje this should fix the pdk build

scotje:
Thanks @gimmyxd !

PR #228, opened 2019-09-23
(PE-26585) Add logic for using correct openssl for FIPS

actowery:
This pr links FIPS boxes to use system openssl in contrast to the binaries that ship with puppet.


highb:
FIPS-140-2-Logo

PR #229, opened 2019-09-27
(packaging) Ship updated bolt gem components

donoghuc:
Pull in gem updates for bolt and bolt-server runtimes.


donoghuc:
Added the `gettext_setup` gem update to a separate PR to see if other projects want to upgrade.

PR #230, opened 2019-09-27
(packaging) Update gettext_setup gem across shared runtimes

donoghuc:
This commit updates the `gettext_setup` gem which is shared by multiple runtime projects.


donoghuc:
This is shared across multiple runtime projects. Do I need to add some conditional logic to allow projects to choose which version? Or is this OK across the board?

PR #231, opened 2019-10-02
(maint) Keep minitar in pdk-runtime at 0.6.1

rodjek:
PDK runtime has to be able to run older versions of Puppet. The mingw
Puppet gems have a runtime dependency of minitar ~> 0.6.1 which is not
satisfied by minitar 0.9.

This somewhat reverts the change in #220, but the default behaviour is
now to include minitar 0.9 (so that this does not affect any other
runtimes) and allowing PDK to opt in to minitar 0.6.1.

I realise that we'll have to update this before the next Puppet release to install both versions of minitar side by side, but that is less pressing at the moment.

### DIFF
--- a/configs/components/rubygem-minitar.rb
+++ b/configs/components/rubygem-minitar.rb
@@ -1,6 +1,16 @@
 component 'rubygem-minitar' do |pkg, settings, platform|
-  pkg.version '0.9'
-  pkg.md5sum '4ab2c278183c9a83f3ad97066c381d84'
+  version = settings[:rubygem_minitar_version] || '0.9'
+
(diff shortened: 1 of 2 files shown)


joshcooper:
I share @donoghuc's question. Does it make more sense to have `puppet-runtime` include the same minitar 0.9 consistently, and then add the additional minitar-0.6.1 to `pdk-vanagon`? I don't understand the tradeoffs well enough.

rodjek:
@donoghuc @joshcooper good question, let me test that out quickly and see

rodjek:
No longer necessary

PR #237, opened 2019-10-16
(PE-27431) Do not build openssl on fips for client tools runtime

justinstoller:
FIPS packages are required to use the system openssl package for SSL
operations, so this commit removes the vendored version in the client-tools
FIPS package


McdonaldSeanp:
I have verified that this works for orch-client and puppet-access.

PR #245, opened 2019-10-31
(PA-2986) Update GCC compiler/linker flags

GabrielNagy:
Harden Linux ELF binaries by compiling with PIE (Position Independent Executables) support, stack canary and full RELRO.

We only do this on platforms that use their default OS toolchain since pl-gcc versions are too old to support these flags.

Dependent on https://github.com/puppetlabs/vanagon/pull/629.

(something similar will need to be done for puppet-agent as well)


GabrielNagy:
Relevant flags are:

```
CPPFLAGS="-D_FORTIFY_SOURCE=2"
CFLAGS="-fstack-protector-strong -fno-plt"
LDFLAGS="-z,relro,-z,now"
```

PR #254, opened 2019-11-26
Add puppet-strings and its dependencies to bolt-runtime

nicklewis:
These are needed for `bolt plan show`.


donoghuc:
@nicklewis @beechtom One of the reasons for managing bolt gem dependencies in puppet-runtime is to keep Gem versions shared between bolt and pe-bolt-server the same. In the past whenever I have added a new gem dependency for bolt I also add it to the shared bolt server runtime project [example](https://github.com/puppetlabs/puppet-runtime/pull/246/files). PR to add puppet-strings + deps: https://github.com/puppetlabs/puppet-runtime/pull/255

PR #259, opened 2019-12-03
(PA-2780) CFPropertyList gem is in the wrong directory

GeorgeMrejea:


gimmyxd:
@GeorgeMrejea Is this solving the issues raised in the ticket? From what i see CFPropertyList is not used by agent-runtime but is declared as a component in puppet-agent: https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/cfpropertylist.rb and the last update of version of made in 2017. We should see if it makes sense to use the rubygem-CFPropertyList defined in puppet-runtime and remove the one from puppet-agent

ciprianbadescu:
I think this should be closed, as work is continued in https://github.com/puppetlabs/puppet-runtime/pull/264

PR #265, opened 2019-12-17
(PE-27977) build client-tools with openssl 1.1.1

sarameisburger:


sarameisburger:
@donoghuc works for me! I forget, do I need to trigger a build or will it trigger on the commit

donoghuc:
Sorry for late response. Yeah everything is automated, runtimes will build and promote into downstream vanagon project branches. You can pop in to #proj-pe-client-tools for updates.

PR #266, opened 2019-12-18
Revert "Merge pull request #265 from sarameisburger/maint"

donoghuc:
This reverts commit 9cdc6b387e651cef73fc98e87476ebabe329d5d3, reversing
changes made to 71291eefead882d17fb6b57860b785dd5db8353f. Openssl 1.1.1 is not compatible with some dependencies in the puppet-code rust project.

### DIFF
--- a/configs/projects/_shared-client-tools-runtime.rb
+++ b/configs/projects/_shared-client-tools-runtime.rb
@@ -7,7 +7,7 @@ unless defined?(proj)
 end
 
 proj.setting(:runtime_project, 'client-tools')
-proj.setting(:openssl_version, '1.1.1')
+proj.setting(:openssl_version, '1.0.2')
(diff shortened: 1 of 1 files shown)


donoghuc:
The puppet-code component `kitchensink` dependency https://github.com/puppetlabs/puppet-code/blob/02de654ddadecbff3f3b878904bbb1ed09a35b5a/Cargo.toml#L15 carries a dependency on `hyperssl` https://github.com/puppetlabs/rust-kitchensink/blob/7bca9591f827840eedc7d17b1d1634665176a842/Cargo.toml#L14 which does not support openssl 1.1.1. 

Note that is claims to support 1.1.0. But for now we should just revert and figure out what we want to do. Luckily client tools appears to ship its own ssl and nothing else links on it.

PR #267, opened 2019-12-18
(PE-27498) Ship net-ssh 5.2.0 with pe-bolt-server

donoghuc:
Previously the agent version of net-ssh was used by bolt-server. This commit updates it to use the same version as bolt and shadow the puppet version to pull in newer features like ed255519 support.


configs/projects/_shared-pe-bolt-server.rb
```diff
@@ -42,6 +42,7 @@
 # TODO: Can runtime projects use these updated versions?
 proj.setting(:rubygem_gettext_version, '3.2.9')
 proj.setting(:rubygem_deep_merge_version, '1.2.1')
+proj.setting(:rubygem_net_ssh_version, '5.2.0')
```
donoghuc: I think it is ok to ship net-ssh in all branches.

branan: Should be

PR #268, opened 2019-12-18
Revert "(PUP-10166) upgrade openssl version"

sarameisburger:
This reverts commit ~888cce2f06326b40f5895d7b46c819689e1500d5.~ 71291eefead882d17fb6b57860b785dd5db8353f

the openssl update has caused some issues in PE smoke tests and we need to pull it out until we can figure out how to fix it/the best path forward

/cc @ciprianbadescu @mihaibuzgau @shrug @joshcooper @melissa

### DIFF
--- a/configs/components/boost.rb
+++ b/configs/components/boost.rb
@@ -46,7 +46,6 @@ component "boost" do |pkg, settings, platform|
   addtl_flags = ""
   gpp = "#{settings[:tools_root]}/bin/g++"
   b2flags = ""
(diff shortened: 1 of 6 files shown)


shrug:
Currently this is breaking postgresql upgrades and I'm not confident we can sort it out before the end of the week, so reverting and hopefully we can sort it out after the first of the year when everyone is back.

sarameisburger:
@joshcooper updated!

PR #269, opened 2019-12-19
(maint) openssl 1.0.2 libs

ciprianbadescu:
besides openssl-1.1.1, deliver also openssl-1.0.2 libs to unblock PE while keeping agent updated

### DIFF
--- a/configs/components/openssl-1.1.1.rb
+++ b/configs/components/openssl-1.1.1.rb
@@ -120,11 +120,7 @@ component 'openssl' do |pkg, settings, platform|
   project_flags = settings[:openssl_extra_configure_flags] || []
   perl_exec = ''
   if platform.is_aix?
-    if platform.name =~ /^aix-6.1/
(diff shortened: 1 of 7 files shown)


configs/components/openssl-1.1.1.rb
```diff
@@ -8,6 +8,10 @@
   # ENVIRONMENT, FLAGS, TARGETS
   #############################
 
+  # make sure openssl-lib compiles first, as we it only installs versioned libs and hopefully will not cause problems
+  # the other way around caused: # error "Inconsistency between crypto.h and cryptlib.c"
+  pkg.build_requires 'openssl-lib' if platform.is_linux?
```
shrug: This is trying to install an 'openssl-lib' package via the package manager - seems like the intention is to have this be dependent on the earlier openssl-lib build?

mihaibuzgau: correct

PR #273, opened 2020-01-06
(maint) bump thor to 1.0.1

mihaibuzgau:


joshcooper:
This change breaks puppet-agent#master builds because it's using [facter-ng-0.0.10](https://github.com/puppetlabs/puppet-agent/blob/e0e9d65ef43858eaa6d0246d1ce2d63f3fd0db2b/configs/components/facter-ng.rb#L2-L3), and that version requires [`thor ~> 0.20.3`](https://github.com/puppetlabs/facter-ng/blob/v0.0.10/facter-ng.gemspec#L41). So the command to `gem install --no-ri --no-rdoc --local facter-ng-0.0.10.gem` fails. See https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/CI%20Goalie/job/platform_puppet-agent_puppet-agent-packaging_daily-master/755/

I'd recommend changing facter-ng so that `thor` Y component can float:

```ruby
s.add_runtime_dependency('thor', [">= 1.0.1", "< 2"])
```

Then we'll need to tag facter-ng, and bump to that version in puppet-agent.

PR #276, opened 2020-01-07
Revert "(maint) bump thor to 1.0.1"

mihaibuzgau:
Reverts puppetlabs/puppet-runtime#273

We're reverting this until facter-ng dependencies are resolved.


joshcooper:
Since a merge commit can introduce changes, it's generally safer to revert the merge commit, using `git revert -m1 e2236fb73e288ca63e3238b294c81031e18a0961`, and it also includes a helpful git commit message.

PR #278, opened 2020-01-08
(PA-3025) enforce fips on fips builds

ciprianbadescu:
if openssl if built with fips options, enforce fips mode by default

### DIFF
--- a/configs/components/openssl-1.0.2.rb
+++ b/configs/components/openssl-1.0.2.rb
@@ -20,7 +20,6 @@ component 'openssl' do |pkg, settings, platform|
     target = platform.architecture == 'x64' ? 'mingw64' : 'mingw'
     cflags = settings[:cflags]
     ldflags = settings[:ldflags]
-    ldflags.slice!("-Wl,--dynamicbase") if platform.name =~ /windowsfips-2012r2/
   elsif platform.is_cross_compiled_linux?
     pkg.environment 'PATH', "/opt/pl-build-tools/bin:$$PATH"
     pkg.environment 'CC', "/opt/pl-build-tools/bin/#{settings[:platform_triple]}-gcc"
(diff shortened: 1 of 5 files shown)


configs/components/openssl-1.0.2.rb
```diff
@@ -84,6 +84,7 @@
     # we can ensure everything is build with the same compiler.
     pkg.apply_patch 'resources/patches/openssl/openssl-mingw-do-not-build-applink.patch'
     pkg.apply_patch 'resources/patches/openssl/openssl-1.0.2t-fix-no-asm-build-on-windows.patch'
+    pkg.apply_patch 'resources/patches/openssl/openssl-enforce-fips-on-fips-mode.patch'
```
gimmyxd: I think we should apply this only for windowsfips, right?

ciprianbadescu: changed code is inside #ifdef OPENSSL_FIPS, should be safe to apply it to all fips openssl builds (https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/o_init.c#L79)

ciprianbadescu:
passing runtime packaging on windows:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/692/

joshcooper:
Since puppet-agent depends on a single puppet-runtime component which contains openssl, how do we ensure the fips puppet-agent builds include fips openssl, while non-fips puppet-agent gets non-fips openssl? Right now it looks like all windows puppet-agent builds are getting the fips openssl, which breaks non-fips puppet-agent when trying to use MD5, the default checksum when sourcing files from puppetserver?

joshcooper:
Ah nevermind, I see there are 2 tarballs. But it seems PEZ creates a PE tarball containing the fips MSI, which breaks PE tests when the MSI is installed on a non-FIPS Windows agent and it tries to compute MD5 file checksum.

joshcooper:
I verified that the last passing puppet-agent#6b729064ecb536e4ad091795061271e58b3c5b3c FIPS MSI behaves as expected:

```
C:\Program Files\Puppet Labs\Puppet\bin>type ..\VERSION
6.12.0.92.g6b729064
C:\Program Files\Puppet Labs\Puppet\bin>puppet apply -e 'md5("")'
md5_dgst.c(75): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
```

And the non-FIPS package works as expected:

```
C:\Program Files\Puppet Labs\Puppet\bin>type ..\VERSION
6.12.0.92.g6b729064
C:\Program Files\Puppet Labs\Puppet\bin>puppet apply -e 'md5("")'
Notice: Compiled catalog for arctic-unease.delivery.puppetlabs.net in environment production in 0.02 seconds
Notice: Applied catalog in 0.01 seconds
```

So the question is how did the FIPS MSI end up in the PEZ tarball?

```
    puppet-enterprise-2019.3.0-rc3-15-g2a044b4-PEZ_cinext-jenkinsmaster-enterprise-prod-1-el-7-x86_64/packages/windows-x86_64/puppet-agent-6.12.0.92.g6b729064-x64.msi
```

ciprianbadescu:
The issue happened during msi signing, I created https://github.com/puppetlabs/packaging/pull/1028 that should fix this problem

PR #282, opened 2020-01-13
(FACT-2272) Added sys-filesystem gem

sebastian-miclea:


sebastian-miclea:
Passing pipeline: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/720/

configs/components/rubygem-ffi.rb
```diff
@@ -83,5 +83,6 @@
     end
   end
 
+
```
ciprianbadescu: nit, not needed empty line

configs/projects/_shared-agent-components.rb
```diff
@@ -48,8 +48,8 @@
 proj.component 'rubygem-gettext'
 proj.component 'rubygem-fast_gettext'
 
+proj.component 'rubygem-ffi'
```
gimmyxd: @sebastian-miclea this should not be build on all platforms on 5.5.x and 6.4.x because we will not use facter-ng on those streams

PR #287, opened 2020-01-17
(PE-28158) Add client-tools/bolt-server master projects

McdonaldSeanp:
This commit adds two new projects: client-tools-runtime-master and
pe-bolt-server-runtime-master.

These new projects will support the new master branch strategy for PE


configs/projects/client-tools-runtime-master.rb
```diff
@@ -0,0 +1,4 @@
+project 'client-tools-runtime-master' do |proj|
+  # Common settings
```
donoghuc: Can we stop building kearney and johnson?

McdonaldSeanp: I don't know. We have platerprise steering next week, I'll ask the folks in there if we can get some clear consensus/communication about the state of all the PE branches.

For now I was just going to leave them

PR #290, opened 2020-01-27
(PA-2994) Add macOS 10.15 Catalina platform

GabrielNagy:
This adds support for building macOS 10.15 runtimes. Boost was bumped since we couldn't build the previous version with clang 11 (version that comes with Catalina Xcode tools).

Bumping boost required an additional Solaris patch, but the code from `solaris-aix-boost-filesystem-unique-path.patch` was rewritten upstream and that patch is not needed anymore.


GabrielNagy:
built runtime: http://builds.delivery.puppetlabs.net/puppet-runtime/172c16c25037f44b87e649deff08ff5654f8254d/artifacts/

PR #294, opened 2020-02-05
(maint) Fix rubygem-ffi on cross compiled builds

GabrielNagy:
Apparently we don't need `libffi` when building the `ffi` gem. The gem will also build `libffi` for us which makes things simpler in the long run.

We still need to patch rubygems to think we're on the destination architecture.

Still testing this...


mihaibuzgau:
I was able to require ffi on solaris 11 sparc:
```
root@sol11-6:~/a# /opt/puppetlabs/puppet/bin/irb
irb(main):001:0> require 'ffi'
=> true
```

PR #295, opened 2020-02-05
Add net-ssh-krb gem to Bolt projects

lucywyman:
The net-ssh-krb gem has been relicensed and can now be shipped as part of the Bolt package. This adds the gem to the bolt-runtime and pe-bolt-server projects at it's latest version, 0.5.1


configs/components/rubygem-ffi.rb
```diff
@@ -1,6 +1,6 @@
 component "rubygem-ffi" do |pkg, settings, platform|
-  pkg.version '1.9.25'
-  pkg.md5sum "e8923807b970643d9e356a65038769ac"
+  pkg.version '1.12.2'
```
donoghuc: Most (all) runtime projects use ffi. Maybe it is not a good idea to bump it. We could ship a different version for bolt but that might not be worth the effort bc I think the failures we are seeing are due to https://github.com/puppetlabs/puppet-runtime/pull/294#issuecomment-582505408

donoghuc:
I would just go ahead an pull in the z gems too. I can think of one case where a bug fix in a z will help bolt: https://github.com/WinRb/WinRM/compare/v2.3.3...master

configs/components/rubygem-multi_json.rb
```diff
@@ -1,6 +1,6 @@
 component "rubygem-multi_json" do |pkg, settings, platform|
-  pkg.version '1.13.1'
```
donoghuc: Unfortunately this is another gem that is shared by most/all projects. I usually have just raised a separate PR for these updates so that teams have time to review it without blocking what I want to get in.

melissa:
There are currently failures building the ffi gem in the puppet-runtime master pipeline, so we should hold off on merging this until those failures are resolved.

lucywyman:
Ok - I'm out next week and would *really* like to get the net-ssh-krb change in. I'm going to pull the gem updates into a separate PR. @donoghuc Do you want to take over the FFI gem update? I think part of this update was hoping that updating the gem would fix those failures, but I don't understand why they are failing.

donoghuc:
Yeah, I will take over the general gem update. I think that nightswatch is working on getting ffi 1.9.25 working again (we should hold off on merging this until that is fixed). Did you end up mirroring the new gem in artifactory?

donoghuc:
Other gem updates https://github.com/puppetlabs/puppet-runtime/pull/297

PR #296, opened 2020-02-05
(maint) Only modify PATH if on Solaris or cross compiling

GabrielNagy:


GabrielNagy:
jenkins please test this with puppet-agent#master on solaris10-SPARC,solaris10-64,solaris11-SPARC,solaris11-64,ubuntu1804-64,fedora28-64

PR #301, opened 2020-02-12
(PA-3113) Patch ruby to allow large environments on Windows

GabrielNagy:
Apply ruby/ruby@490fbc4 (which is only in ruby 2.7 and higher) to fix an issue with large environment blocks on Windows.


GabrielNagy:
Confirmed this works by building a runtime and setting tons of variables in ruby:

```ruby
100_000.times { ENV["#{rand(99999)}"] = "salam" }
```

In contrast, running this with the ruby that comes with an existing agent produces something along the following:

```
irb(main):003:0> 10_000.times { ENV["#{rand(99999)}"] = "salam" }
Traceback (most recent call last):
        5: from C:/Program Files/Puppet Labs/Puppet/puppet/bin/irb.bat:19:in `<main>'
        4: from (irb):3
        3: from (irb):3:in `times'
        2: from (irb):3:in `block in irb_binding'
        1: from (irb):3:in `[]='
Errno::EINVAL (Invalid argument - ruby_setenv(66739))
irb(main):004:0> ENV.count
=> 2600
```

PR #313, opened 2020-04-01
(packaging) Bump shared rubygem component versions to latest

lucywyman:
This updates rubygem component versions to their latest for gems that
are shared across multiple projects.

Specifically the @puppetlabs/installer-and-management team includes:
bcrypt_pbkdf
ffi
gettext-setup
locale
multi_json

The puppet-agent package includes:
concurrent-ruby
ffi
gettext-setup
locale
multi_json

The @puppetlabs/pdk package includes:
ffi
gettext-setup

### DIFF
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -41,6 +41,8 @@ component "rubygem-ffi" do |pkg, settings, platform|
     pkg.install_file "#{settings[:tools_root]}/#{settings[:platform_triple]}/sysroot/usr/lib/libffi.so.5.0.10", "#{settings[:libdir]}/libffi.so"
   elsif platform.name =~ /solaris-11-i386/
     pkg.install_file "/usr/lib/libffi.so.5.0.10", "#{settings[:libdir]}/libffi.so"
+  elsif platform.name =~ /solaris-10-i386/
+    pkg.install_file "/opt/csw/lib/libffi.so.6", "#{settings[:libdir]}/libffi.so.6"
   end
 
   if platform.name =~ /el-5-x86_64/
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -1,6 +1,6 @@
 component "rubygem-ffi" do |pkg, settings, platform|
-  pkg.version '1.9.25'
-  pkg.md5sum "e8923807b970643d9e356a65038769ac"
+  pkg.version '1.12.2'
+  pkg.md5sum "28dc3d1294a04b728d24ba025e331b13"
 
   instance_eval File.read('configs/components/_base-rubygem.rb')
 
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -7,6 +7,9 @@ component "rubygem-ffi" do |pkg, settings, platform|
   # Windows versions of the FFI gem have custom filenames, so we overwite the
   # defaults that _base-rubygem provides here, just for Windows.
   if platform.is_windows?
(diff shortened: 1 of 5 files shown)


configs/components/rubygem-concurrent-ruby.rb
```diff
@@ -1,6 +1,6 @@
 component 'rubygem-concurrent-ruby' do |pkg, settings, platform|
-  pkg.version '1.1.5'
-  pkg.md5sum '4409c2d6925d8448cb34a947eacaa29b'
+  pkg.version '1.1.6'
```
donoghuc, line 2: Please do not take 1.1.6 https://github.com/ruby-concurrency/concurrent-ruby/pull/856#issuecomment-604124810 will be needed. 1.1.5 still works. (this is for ace-server)

configs/components/rubygem-ffi.rb
```diff
@@ -1,6 +1,6 @@
 component "rubygem-ffi" do |pkg, settings, platform|
-  pkg.version '1.9.25'
-  pkg.md5sum "e8923807b970643d9e356a65038769ac"
+  pkg.version '1.12.2'
+  pkg.md5sum "28dc3d1294a04b728d24ba025e331b13"
```
scotje, line 3: I know we've had some challenges with "ffi" versions in the past in PDK, I'll see if I can make an ad-hoc build with this PR and validate this works.

donoghuc: Same with bolt on OSX. might be worth seeing that as well with this ffi

scotje: @lucywyman we need to update the md5 sums for the windows version of the `ffi` gem a little further down in the file.

GabrielNagy:
Was able to successfully build puppet-runtime for agent-master, also triggering a puppet-agent build with these changes.

GabrielNagy:
We're encountering some valid failures with the ffi gem installation on Solaris, so please do not merge this just yet.

GabrielNagy:
There are also vaild pxp-agent test failures on Windows:

```
18:12:01 Minitest::Assertion: Successful action expected to have no output on stderr.
18:12:01 Expected "[DEPRECATION] Struct layout is already defined for class Windows::ServiceStructs::SERVICE_STATUS_PROCESS. Redefinition as in C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/gems/2.5.0/gems/win32-service-0.8.8/lib/win32/windows/structs.rb:67:in `<class:SERVICE_STATUS_PROCESS>' will be disallowed in ffi-2.0.\n" to be nil.
```
Traced that to https://github.com/chef/win32-service/issues/70

We should hold back the FFI gem for Windows.

configs/components/rubygem-ffi.rb
GabrielNagy: @scotje Apologies, I dropped your commit since newer ffi breaks things on Windows (see comments below), so there's no need to update the checksums.

scotje: No worries!

scotje:
Still working to test a PDK build, the `pdk` gem itself was pinned to `ffi` 1.9.x so I've expanded the range it will accept and am testing with updated runtime now.

lucywyman:
@donoghuc One more lookover from skeletor? @actowery Would you be willing to look this over and verify the updates to packages that the installer team consumes will work?

donoghuc:
If you can get bolt builds on all the osx platforms them i'm good. We only build bolt-server for master platforms. i was just noting that I had previously had problems building ffi > 1.9.25 on osx platforms.

shaigy:
👍

PR #321, opened 2020-04-13
(PA-3179) fix ruby download url

gimmyxd:
Update the gsub on ruby version to handle multiple digits
z version

```
# before
>> "2.4.10".gsub(/(\d)\.(\d)\.(\d)/, '\1.\2') #=> "2.40"

# now
>> "2.4.10".gsub(/(\d)\.(\d)\.(\d+)/, '\1.\2') #=> "2.4"
>> "2.4.1".gsub(/(\d)\.(\d)\.(\d+)/, '\1.\2') #=> "2.4"
```

### DIFF
--- a/configs/components/_base-ruby.rb
+++ b/configs/components/_base-ruby.rb
@@ -5,7 +5,7 @@
 # Condensed version, e.g. '2.4.3' -> '243'
 ruby_version_condensed = pkg.get_version.tr('.', '')
 # Y version, e.g. '2.4.3' -> '2.4'
-ruby_version_y = pkg.get_version.gsub(/(\d)\.(\d)\.(\d)/, '\1.\2')
+ruby_version_y = pkg.get_version.gsub(/(\d)\.(\d)\.(\d+)/, '\1.\2')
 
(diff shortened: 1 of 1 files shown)


gimmyxd:
seems like the azure job hase an extra `:` for the runtime url
```
2020-04-13T06:16:08.6275660Z [91mhttps://artifactory.delivery.puppetlabs.net/artifactory/rpm__local/development/puppet-runtime/202004120/el-7-x86_64/agent-runtime-master-202004120.el-7-x86_64.tar.gz:
2020-04-13T06:16:08.6278361Z 2020-04-13 06:16:07 ERROR 404: Not Found.
``

GabrielNagy:
@gimmyxd maybe that comes from the error message, the URL is not available since runtime was tagged but not built (because of the issue fixed by this PR)

gimmyxd:
> @gimmyxd maybe that comes from the error message, the URL is not available since runtime was tagged but not built (because of the issue fixed by this PR)

yes, seems like that, my assumption with `:` was wrong

PR #324, opened 2020-04-15
Use linux-aarch64 openssl target on aarch64

traylenator:
Using the platform el-8-aarch64 openssl build target was being
incorrectly specified as linux-x86_64. Use the correct
openssl linux-aarch64 instead.

### DIFF
--- a/configs/components/openssl-1.0.2.rb
+++ b/configs/components/openssl-1.0.2.rb
@@ -59,6 +59,8 @@ component 'openssl' do |pkg, settings, platform|
     if platform.architecture =~ /86$/
       target = 'linux-elf'
       sslflags = '386'
+    elsif platform.architecture =~ /aarch64$/
+      target = 'linux-aarch64'
     elsif platform.architecture =~ /64$/
       target = 'linux-x86_64'
     end
--- a/configs/components/openssl-1.1.1.rb
+++ b/configs/components/openssl-1.1.1.rb
@@ -72,6 +72,8 @@ component 'openssl' do |pkg, settings, platform|
     if platform.architecture =~ /86$/
       target = 'linux-elf'
       sslflags = '386'
+    elsif platform.architecture =~ /aarch64$/
+      target = 'linux-aarch64'
     elsif platform.architecture =~ /64$/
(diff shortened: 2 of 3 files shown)


configs/components/openssl-1.0.2.rb
```diff
@@ -59,6 +59,8 @@
     if platform.architecture =~ /86$/
       target = 'linux-elf'
       sslflags = '386'
+    elsif platform.architecture =~ /aarch64$/
```
gimmyxd, line 62: @traylenator isn't this caught in the above condition?
https://github.com/puppetlabs/puppet-runtime/pull/324/files#diff-4a81b1af0a76a0671cb08cfd075dff48R23-R37

```ruby
elsif platform.is_cross_compiled_linux?
    pkg.environment 'PATH', "/opt/pl-build-tools/bin:$$PATH"
    pkg.environment 'CC', "/opt/pl-build-tools/bin/#{settings[:platform_triple]}-gcc"

    cflags = "#{settings[:cflags]} -fPIC"
    if platform.architecture =~ /aarch/
      # OpenSSL fails to work on aarch unless we turn down the compiler optimization.
      # See PA-2135 for details
      cflags += " -O2"
    end

    ldflags = "-Wl,-rpath=/opt/pl-build-tools/#{settings[:platform_triple]}/lib -Wl,-rpath=#{settings[:libdir]} -L/opt/pl-build-tools/#{settings[:platform_triple]}/lib"
    target = if platform.architecture == 'aarch64'
                'linux-aarch64'
              elsif platform.name =~ /debian-8-arm/
                'linux-armv4'
              elsif platform.architecture =~ /ppc64le|ppc64el/ # Little-endian
                'linux-ppc64le'
              elsif platform.architecture =~ /ppc64/ # Big-endian
                'linux-ppc64'
              end
```

traylenator: Only if cross compiling rather than native.

GabrielNagy: This makes sense. We didn't do this before since we were always cross-compiling.

PR #327, opened 2020-04-22
(maint) Bump net-ssh gem to 6.0.2 for Bolt

lucywyman:
The net-ssh gem release 6.0.2 which includes several improvements
relevant to Bolt, including support for the OpenSSH
`StrictHostKeyChecking` option, performance improvements, and
curve25519sha256 key support.

### DIFF
--- a/configs/components/rubygem-net-ssh.rb
+++ b/configs/components/rubygem-net-ssh.rb
@@ -4,6 +4,8 @@ component "rubygem-net-ssh" do |pkg, settings, platform|
   pkg.version version
 
   case version
+  when "6.0.2"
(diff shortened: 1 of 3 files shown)


donoghuc:
Do you have time to try to build bolt-server? We are super busy rn

donoghuc:
I was hoping you would add this to bolt-server too, https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/_shared-pe-bolt-server.rb#L45 we want to be on the same version as bolt.

lucywyman:
Well, it looks like we're going to wait to bump this anyway: https://github.com/net-ssh/net-ssh/issues/728 https://github.com/puppetlabs/bolt/pull/1758

PR #329, opened 2020-04-28
(maint) Enable Docker Buildkit

Iristyle:
- Remove wget dependency from container (leverages docker ADD from URL which works better with buildkit caching)
- Disable global linting in favor of inline exclusions where necessary
- Upgrade docker-compose to 1.25.5
- Replace unnecessary ENV vars with ARGs
- Enable buildkit in Azure / WSL2
- Follow standard Docker practice and COPY the Dockerfile to /


docker/puppet-runtime/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM centos:8
```
Iristyle, line 1: This didn't fix Travis :(

Iristyle: Will add in Buildkit on Travis in a separate PR.... though this container is so simple it's probably not a big deal either way.

PR #333, opened 2020-05-07
(maint) Add x25519 gem dependency to bolt-runtime

beechtom:
This adds the `x25519` gem dependency to bolt packaging, except on
Windows platforms, 32-bit platforms, and el6.

### DIFF
--- a/configs/projects/bolt-runtime.rb
+++ b/configs/projects/bolt-runtime.rb
@@ -122,6 +122,12 @@ project 'bolt-runtime' do |proj|
     proj.component 'rubygem-ed25519'
   end
 
+  # Building native extensions for the x25519 gem is only supported
+  # on 64-bit, non-Windows systems.
+  unless platform.is_windows? || platform.dist == 'el6' || platform.architecture == 'i386'
+    proj.component 'rubygem-x25519'
+  end
+
   # Puppet dependencies
   proj.component 'rubygem-hocon'
   proj.component 'rubygem-deep_merge'
--- a/configs/components/rubygem-x25519.rb
+++ b/configs/components/rubygem-x25519.rb
@@ -1,0 +1,6 @@
+component 'rubygem-x25519' do |pkg, settings, platform|
+  pkg.version '1.0.8'
+  pkg.md5sum '6d6fd1c422a6d1e370ea6aa2ee356982'
+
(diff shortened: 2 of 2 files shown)


configs/projects/bolt-runtime.rb
```diff
@@ -122,6 +122,12 @@
     proj.component 'rubygem-ed25519'
   end
 
+  # Building native extensions for the x25519 gem is only supported
+  # on 64-bit, non-Windows systems.
+  unless platform.is_windows? || %w[el6 ubuntu16.04 debian8].include?(platform.dist)
```
donoghuc, line 127: Can you add this logic to the bolt-server project as well? (just for ubuntu and el6 bc we only build for master platforms)

beechtom: `x25519` isn't compatible with jruby, so the team thought it shouldn't be included in bolt-server.

donoghuc: Bolt server uses mri ruby? Is that like a long term "if we move bolt-server to the JVM" thing?

beechtom: Yes. If the Skeletor team would like this added to bolt-server anyways I can update the PR.

donoghuc: meh, probs not worth it. I dont know the timeline for moving to JVM which means it probably not soon.

configs/projects/_shared-pe-bolt-server.rb
```diff
@@ -74,6 +74,12 @@
 # R10k dependencies
 proj.component('rubygem-gettext-setup')
 
+# Building native extensions for the x25519 gem isn't supported on
+# el6 or 32-bit systems
+unless %w[el6 ubuntu16.04].include?(platform.dist)
```
donoghuc, line 79: So it builds on amd64? we only build bolt-server for 64 bit ubuntu. So you can remove that here and only skip el6. So bolt wont support either arch on ubuntu 16.04? Seems like maybe there should be a way to differentiate between the two?

donoghuc:
OK. i guess we are not shipping this with PE (so you can ignore adding it to bolt-server runtime). Sorry for the churn. The note about not shipping for both the 32 and 64 bit ubuntu/debian for bolt is still an open question though.

PR #336, opened 2020-05-08
(maint) Enable Docker buildkit linux

Iristyle:
- Switches Travis to Ubuntu 20.04 builder that has Docker 19.03 pre-installed, so that removes ~30s of setup time from builds
- Enable Docker buildkit (for this project it's mostly an improvement for local caching)
- Buildkit in Travis doesn't seem to work right pulling CentOS with the `--pull` switch for `docker build`, so manually download the image prior to building (based on https://github.com/puppetlabs/puppet-agent/pull/1906 / https://github.com/puppetlabs/puppet-agent/pull/1907)
- Separate out Travis operations so that logs have timings for lint, build and test


Iristyle:
make lint: 2.84s
make build: 42.56s
make test: 10.96s

PR #341, opened 2020-05-26
(HC-105) bump hocon gem to 1.3.1

pcarlisle:
This fixes corruption to internal state that happened when using the hocon gem
concurrently in multiple threads.


Magisus:
Looks good, gonna do a test build.

PR #342, opened 2020-06-02
(PA-3250) remove env check for OPENSSL_FIPS

ciprianbadescu:
OPENSSL_FIPS env variable was used inside openssl 1.0.2 application
to throw error if openssl was not compiled in fips mode.

After move to openssl 1.1.0/1.1.1 branch (that is not supporting FIPS mode)
the check was change to dirrectly throw error if OPENSSL_FIPS is set.

After this commit, the check is removed for redhat7 vendored openssl fips


ciprianbadescu:
local build was successful:
```
cp bill-of-materials output/agent-runtime-master-202005260.redhatfips-7-x86_64-bill-of-materials ||:
cp agent-runtime-master-202005260.redhatfips-7-x86_64.tar.gz output
sha1sum output/agent-runtime-master-202005260.redhatfips-7-x86_64.tar.gz > output/agent-runtime-master-202005260.redhatfips-7-x86_64.tar.gz.sha1
```

PR #348, opened 2020-06-17
(packaging) Bump Bolt dependencies to latest versions

lucywyman:
This updates gems used only by Bolt and PE Bolt Server packages to their
latest versions.


lucywyman:
The powershell failure doesn't seem related...

IrimieBogdan:
@lucywyman please consider upgrading to Facter [4.0.27](https://rubygems.org/gems/facter/versions/4.0.27) it fixes a race condition that was affecting Bolt https://tickets.puppetlabs.com/browse/FACT-2678

PR #349, opened 2020-06-17
(packaging) Bump shared rubygem versions

lucywyman:
This updates rubygems used by multiple projects to their latest versions.

@puppetlabs/night-s-watch Need approval for bumping all three gems

@puppetlabs/installer-and-management Need approval for `ffi`

@puppetlabs/pdk Need approval for `ffi`

### DIFF
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -1,6 +1,6 @@
 component "rubygem-ffi" do |pkg, settings, platform|
-  pkg.version '1.12.2'
-  pkg.md5sum "28dc3d1294a04b728d24ba025e331b13"
+  pkg.version '1.13.1'
+  pkg.md5sum "569d4561359cb3c8d8360d26bd1be4ed"
 
   instance_eval File.read('configs/components/_base-rubygem.rb')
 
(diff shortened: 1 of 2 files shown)


scotje:
Going to build a `pdk-runtime` package from this and then run it through our tests.

gimmyxd:
Jenkins please test this with puppet-agent master on aix61-POWERfa,solaris10-64a,windows2012r2-64a

configs/components/rubygem-concurrent-ruby.rb
```diff
@@ -1,6 +1,6 @@
 component 'rubygem-concurrent-ruby' do |pkg, settings, platform|
-  pkg.version '1.1.5'
-  pkg.md5sum '4409c2d6925d8448cb34a947eacaa29b'
+  pkg.version '1.1.6'
```
donoghuc, line 2: I think i tried bumping this before and hit https://github.com/ruby-concurrency/concurrent-ruby/pull/856#issuecomment-604124810 I verified that patch stops the segfault, but it has not been merged/released yet. I think we should wait for that to be addressed.

lucywyman: Removed!

lucywyman:
Just as a heads up, PDK had trouble verifying a build due to a Litmus dependency failing to compile. We're waiting on that issue to get resolved so that they can ensure this works.

PR #353, opened 2020-06-30
(maint) Try to install libtinfo ubuntu 1804 dep

donoghuc:
This commit attempts to explicitly install a dependency for libtinfo. However this appears to not be a complete solution.

### DIFF
--- a/configs/platforms/ubuntu-18.04-amd64.rb
+++ b/configs/platforms/ubuntu-18.04-amd64.rb
@@ -10,6 +10,7 @@ platform "ubuntu-18.04-amd64" do |plat|
 
   packages = [
     "libbz2-dev",
+    "libtinfo-dev",
     "libreadline-dev",
     "libselinux1-dev",
     "make",


donoghuc:
Raising this bc we are seeing:
```
08:42:06 The following packages have unmet dependencies:
08:42:06  libreadline-dev : Depends: libtinfo-dev but it is not going to be installed
08:42:06 E: Unable to correct problems, you have held broken packages.
```
With this change we now see
```
The following packages have unmet dependencies:
 libtinfo-dev : Depends: libtinfo5 (= 6.1-1ubuntu1) but 6.1-1ubuntu1.18.04 is to be installed
```

donoghuc:
The root cause of this issue appears to be a change in facter behaviour that resulted in the `/etc/apt/sources.list.d/bionic-updates.list` file not being present on 1804 images on vmpooler. https://github.com/puppetlabs/puppetlabs-packer/blob/IMAGES-1217/manifests/modules/packer/manifests/vsphere/repos.pp#L54-L64

PR #354, opened 2020-07-03
(maint) Fix ffi install on cross-compiled

gimmyxd:
FFI 1.13 forced the minimum required ruby version to ~> 2.3
In order to be able to install the gem using pl-ruby(2.1.9)
we need to remove the required ruby version check


gimmyxd:
image

PR #356, opened 2020-07-13
(maint) Add bundle path to bolt-runtime

lucywyman:
In order to run a rake task to generate the Bolt cmdlets Powershell
module in bolt-vanagon we need to use bundler to install gems at build time.
This adds the platform-dependent bundle-path setting to bolt-runtime so that we
can easily call bundler from the Bolt component to create a ruby environment
and run the rake task.


donoghuc:
Should this be managed in the bolt-vanagon project? I think in general gems that are needed only for building (that are never shipped) are not managed like gem components that are part of final builds.

lucywyman:
You're totally right, I got stuck on autopilot.

PR #359, opened 2020-07-27
(PA-3295) Ubuntu 14.04 removal from puppet runtime

Dorin-Pleava:
This PR removes Ubuntu 14.04 platforms and cleanups checks.


Dorin-Pleava:
Creating a new PR with all puppet-runtime removals to avoid conflicts.

PR #360, opened 2020-07-28
(PA-3347) Add ruby 2.7 support for cross-compiled platforms

GabrielNagy:
Removed some patches, and added way more to account for using old host Ruby versions to build/install Ruby.

Also, switch to using system provided host rubies since Ruby tries to require openssl when installing, which fails on pl-ruby since it wasn't built with openssl support.

PR #361, opened 2020-07-30
(PA-3350) Added scanf gem to puppet-runtime

Dorin-Pleava:


configs/projects/_shared-agent-components.rb
```diff
@@ -51,6 +51,7 @@
 proj.component 'rubygem-locale'
 proj.component 'rubygem-gettext'
 proj.component 'rubygem-fast_gettext'
+proj.component 'rubygem-scanf'
```
GabrielNagy, line 54: This should be only for agent-runtime-next.rb which I'm adding in https://github.com/puppetlabs/puppet-runtime/pull/360.

So you can just add the component here then I'll rebase my PR.

PR #363, opened 2020-08-05
(maint) Pin back ffi on Windows

GabrielNagy:
We accidentally bumped ffi for Windows when merging https://github.com/puppetlabs/puppet-runtime/pull/362.

I think we only want this for the `agent-runtime-main` project, where we ship Ruby 2.7, which does not work with older versions of FFI.


GabrielNagy:
Ran some quick checks with `bundle exec inspect` with different projects and Windows combinations and things look good.

PR #364, opened 2020-08-05
(PUP-10590) Remove win32 gems from puppet-agent dependency

luchihoratiu:
Since the `win32-dir`,`win32-process`,`win32-security`,`win32-service` gems will no longer be used in Puppet 7, this commit moves them from the puppet runtime shared agent components to the specific branches that will continue using (5.5.x and 6.x) to avoid unnecessary packaging.

Depends on [PUP-7445](https://github.com/puppetlabs/puppet/pull/8204), [PUP-5758](https://github.com/puppetlabs/puppet/pull/8194) and [PUP-6184](https://github.com/puppetlabs/puppet/pull/8258).

### DIFF
--- a/configs/projects/_shared-agent-components.rb
+++ b/configs/projects/_shared-agent-components.rb
@@ -59,7 +59,6 @@ end
 
 if platform.is_windows?
   proj.component 'rubygem-win32-dir'
-  proj.component 'rubygem-win32-process'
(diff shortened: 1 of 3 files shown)


GabrielNagy:
Ran an ad-hoc, looks good:
image

PR #365, opened 2020-08-05
(PUP-7446) Remove win32-service gem next agent dependency

luchihoratiu:
Since the `win32-service` gem will no longer be used in Puppet 7, this
commit moves it from the puppet runtime shared agent components to the
specific branches that will continue using (5.5.x and master) to avoid
unnecessary packaging.

Depends on [PUP-5758](https://github.com/puppetlabs/puppet/pull/8194)


luchihoratiu:
Commit moved in #364

PR #366, opened 2020-08-14
(PA-3362) Fix ffi gem compilation on Solaris 11 SPARC

GabrielNagy:
Previously we've been using the system libffi to build the ffi gem on Solaris 11 SPARC.

This caused all FFI calls to fail with the following message:

`[BUG] Bus Error at 0x033e66d3`

It appears that we can fix this by forcing a static build of the ffi gem native extension. To accomplish this we do the following:

- avoid installing libffi/libffi_dev during the provisioning step
- force compilation to use the GNU ld linker (the simplest way to accomplish this is to temporarily move the Solaris linker somewhere else)
- append a couple of paths to LDFLAGS so ld knows where to find system libraries

### DIFF
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -52,9 +52,7 @@ component "rubygem-ffi" do |pkg, settings, platform|
     pkg.environment "PATH", "/opt/pl-build-tools/bin:/opt/csw/bin:$$PATH"
   end
 
-  if platform.name =~ /solaris-11-sparc/
-    pkg.install_file "#{settings[:tools_root]}/#{settings[:platform_triple]}/sysroot/usr/lib/libffi.so.5.0.10", "#{settings[:libdir]}/libffi.so"
-  elsif platform.name =~ /solaris-11-i386/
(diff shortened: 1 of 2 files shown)


GabrielNagy:
jenkins please test this with puppet-agent#main on solaris11-SPARC

GabrielNagy:
jenkins please test this with puppet-agent#master on solaris11-SPARC

configs/components/ruby-2.7.1.rb
```diff
@@ -192,6 +192,8 @@
       # will remove that entry
       # Matches both endians
       rbconfig_changes["DLDFLAGS"] = "-Wl,-rpath=/opt/puppetlabs/puppet/lib -L/opt/puppetlabs/puppet/lib  -Wl,-rpath,/opt/puppetlabs/puppet/lib"
+    elsif platform.name =~ /solaris-11-sparc/
```
GabrielNagy, line 195: oops, these aren't needed, will remove them after PR testing finishes

GabrielNagy:
jenkins please test this with puppet-agent#main on solaris11-SPARC

PR #369, opened 2020-08-26
(maint) Bump puppet rubygem to 6.18.0

lucywyman:
Bumps the puppet rubygem to latest 6.18.0

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -1,6 +1,6 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
-  pkg.version '6.16.0'
-  pkg.md5sum 'ed0c92a2bbeb5247bff7e33fb85af873'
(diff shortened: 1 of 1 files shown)


donoghuc:
Why is this bolt only?

donoghuc:
We will need 6.18.0 in bolt-server runtime.

lucywyman:
I wasn't sure if other teams were ready to pick up the latest Puppet, so wanted to make it optional. If everyone's ready to upgrade I can just update it. We have some work blocked on this in Bolt, so I'd like to get it in by Monday so we can release new features.

donoghuc:
I think anybody that needs the puppet gem just needs latest bolt. I cant think of a reason not to bump it everywhere.

lucywyman:
@donoghuc Just FYI I think we're gonna wait to merge this until the Bolt bits are ready to go, since they're mutually dependent and we need to do some rebasing since the PRs are so old.

PR #371, opened 2020-09-14
(maint) Bump Bolt rubygem components

beechtom:
Bumps the version of Bolt rubygem components.


lucywyman:
While we're here can we update all our gems? Don't have to if you don't have time, just the longer we wait the more work it'll be.

configs/components/rubygem-logging.rb
```diff
@@ -1,6 +1,6 @@
 component 'rubygem-logging' do |pkg, settings, platform|
-  pkg.version '2.2.2'
-  pkg.md5sum '13c12082ef0b69a639ed7ca318e7f514'
+  pkg.version '2.3.0'
```
donoghuc, line 2: I think this release included a fix for this. We will want to *not* patch anymore https://github.com/puppetlabs/puppet-runtime/tree/master/resources/patches/rubygem-logging

donoghuc (changes requested):

donoghuc (approved):

lucywyman (approved):

lucywyman:
Verified these only touch pe-bolt-server and bolt, and built + installed :+1:

PR #375, opened 2020-09-23
(PDK-1673) Update PDK OpenSSL to 1.1.1

underscorgan:


ciprianbadescu (approved):

scotje:
Building this now and then going to run it through PDK CI, but should be good.

scotje:
I think the only issue here is that we are currently still building Ruby 2.1.9 for PDK and Ruby < 2.4 is not compatible with OpenSSL 1.1.x, but we were planning to drop that as part of the Puppet 7 addition anyway, so this might just have to wait until that other work is done.

underscorgan:
Hey @scotje ! Just wanted to follow back up on this since puppet7 is out.

Thanks!

underscorgan:
@jpogran can you take a look at this? Thank you!!

jpogran:
We've stopped shipping ruby 2.1 in the PDK 2.0 release, so this should be good to merge, but I will check with @scotje just in case I missed some context

PR #383, opened 2020-10-21
Bump gem versions to latest

lucywyman:
This bumps several gems used by Bolt to their latest versions in the
current Bolt Gemfile.lock. This helps us stay up-to-date with updates
from gems, and catch any issues with those updated versions in packaged
Bolt early.

CC @puppetlabs/night-s-watch Looks like the agent uses concurrent-ruby and multi_json

@puppetlabs/installer-and-management uses multi_json

And @puppetlabs/skeletor shares all of these


lucywyman:
Haha closing in favor of #411

PR #387, opened 2020-11-10
(FIXUP) Change `fstack-protector` rbconfig for Ruby 2.7.1 on RHEL7

scotje:
For whatever reason, the flag support detection in Ruby >2.6.0 is improperly including `-fstack-protector-strong` in `LDFLAGS` even on GCC versions that don't support it.

Not 100% sure this is the right/best way to fix this (just copied what was being done above for `sles-12-ppc64le`) but this does seem to fix the issues I was having compiling native gems with `pdk-runtime` on RHEL7.

### DIFF
--- a/configs/components/ruby-2.7.1.rb
+++ b/configs/components/ruby-2.7.1.rb
@@ -198,6 +198,9 @@ component 'ruby-2.7.1' do |pkg, settings, platform|
     end
   elsif platform.is_windows?
     rbconfig_changes["CC"] = "x86_64-w64-mingw32-gcc"
+  elsif platform.name =~ /el-7-x86_64/
+    # EL 7 GCC (4.8.x) does not support -fstack-protector-strong
+    rbconfig_changes["LDFLAGS"] = "-L. -Wl,-rpath=/opt/puppetlabs/puppet/lib -fstack-protector -rdynamic -Wl,-export-dynamic -L/opt/puppetlabs/puppet/lib"
   end
 
   pkg.add_source("file://resources/files/ruby_vendor_gems/operating_system.rb")


scotje:
Need this resolved for PDK / Platform 7 release.

GabrielNagy:
I think I know what's happening here:

CentOS/RedHat 7 machines on vmpooler come with gcc installed (4.8.5), which is actually a bit newer than the gcc shipped with pl-build-tools (4.8.2).

Native GCC supports the `-fstack-protector-strong` flag whereas the GCC from pl-build-tools does not. What happens is, Ruby is configured to build with the native gcc, which is why it reports that `-fstack-protector-strong` is supported. Which is also why building Ruby doesn't fail as well.

However, gems use the pl-build-tools GCC when building native extensions, but the LDFLAGS are inherited from the Ruby build... so using `-fstack-protector-strong` fails.

configs/components/ruby-2.7.1.rb
```diff
@@ -198,6 +198,9 @@
     end
   elsif platform.is_windows?
     rbconfig_changes["CC"] = "x86_64-w64-mingw32-gcc"
+  elsif platform.name =~ /el-7-x86_64/
+    # EL 7 GCC (4.8.x) does not support -fstack-protector-strong
+    rbconfig_changes["LDFLAGS"] = "-L. -Wl,-rpath=/opt/puppetlabs/puppet/lib -fstack-protector -rdynamic -Wl,-export-dynamic -L/opt/puppetlabs/puppet/lib"
```
GabrielNagy, line 203: I wonder if we can avoid overriding LDFLAGS (especially since it's project-based, pdk should use `/opt/puppetlabs/pdk` instead of `/opt/puppetlabs/puppet`). Maybe we can get away with setting `rbconfig_changes["CC"] = "/usr/bin/gcc"`, so gems with native extensions use the system Ruby which does support the flag.

scotje:
Hmmm, let me try just bypassing the `pl-build-tools` provided GCC in `pdk-vanagon` for RHEL7 and see if that also resolves this. If we aren't using it to build the runtime, I don't really see a reason to use it in `pdk-vanagon` either.

scotje:
Going to close this, switching pdk-vanagon to not use `pl-build-tools` for EL7 seems to have fixed things. Thanks @GabrielNagy !

PR #408, opened 2021-01-04
(maint) Bump shared rubygems for bolt, pe-bolt-server

beechtom:
Part of https://github.com/puppetlabs/bolt/pull/2496


configs/components/rubygem-aws-partitions.rb
```diff
@@ -1,6 +1,6 @@
 component "rubygem-aws-partitions" do |pkg, settings, platform|
-  pkg.version "1.371.0"
-  pkg.md5sum "6b70b220604890f584a7b3be2bd6a971"
+  pkg.version "1.407.0"
```
lucywyman, line 2: Some of these can be further updated.

lucywyman (changes requested):

PR #412, opened 2021-01-14
(PA-3532) Add support for el-8-ppc64le

GabrielNagy:


configs/components/openssl-1.0.2.rb
```diff
@@ -61,6 +61,8 @@
       sslflags = '386'
     elsif platform.architecture =~ /aarch64$/
       target = 'linux-aarch64'
+    elsif platform.architecture =~ /ppc64le|ppc64el/ # Little-endian
+      target = 'linux-ppc64le'
```
gimmyxd, line 65: I don't think we should modify openssl 1.0.2 as we don't add support for 5.5.x stream

configs/components/openssl-1.1.1.rb
```diff
@@ -74,6 +74,8 @@
       sslflags = '386'
     elsif platform.architecture =~ /aarch64$/
       target = 'linux-aarch64'
+    elsif platform.architecture =~ /ppc64le|ppc64el/ # Little-endian
+      target = 'linux-ppc64le'
```
gimmyxd, line 78: Won't this impact the other ppc64 platforms, that are cross compiled?

GabrielNagy: Cross compiled platforms should be caught on line 29

gimmyxd (approved):

PR #416, opened 2021-01-21
(PA-3541) Compile curl with CA path

GabrielNagy:
The curl command is currently compiled without the `--with-ca-path` flag.

This breaks the ability to configure SSL trust for downloading files using curl separately from SSL trust for the Puppet server/agent
applications.

When we configure and build curl, we should pass the `--with-ca-path /opt/puppetlabs/puppet/ssl/certs` flag.

Curl does not provide a way to see the CApath, but overriding the CAfile should print both flags.

Prior to this commit:
```sh-session
> /opt/puppetlabs/puppet/bin/curl --cacert a https://google.com
curl: (77) error setting certificate verify locations:
  CAfile: a
  CApath: none
```

With this commit:
```sh-session
> /opt/puppetlabs/puppet/bin/curl --cacert a https://google.com
curl: (77) error setting certificate verify locations:
  CAfile: a
  CApath: /opt/puppetlabs/puppet/ssl/certs
```

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -47,6 +47,7 @@ component 'curl' do |pkg, settings, platform|
         --disable-ldap \
         --disable-ldaps \
         --with-ca-bundle=#{settings[:prefix]}/ssl/cert.pem \
+        --with-ca-path=#{settings[:prefix]}/ssl/certs \
         CFLAGS='#{settings[:cflags]}' \
         #{settings[:host]}"]
   end


reidmv (approved):
LGTM. Basically, this just makes curl look at the already-present <prefix>/ssl/certs directory. Per the documentation for curl's config options themselves, and also per the man page for c_rehash, <prefix>/ssl/certs is a reasonable value for this.

Directory naming hints from those materials:

```
--with-ca-path=DIRECTORY
                          Path to a directory containing CA certificates
                          stored individually, with their filenames in a hash
                          format. This option can be used with OpenSSL, GnuTLS
                          and PolarSSL backends. Refer to OpenSSL c_rehash for
                          details. (example: /etc/certificates)
```

> If any directories are named on the command line, then those are processed in turn. If not, then the SSL_CERT_DIR environment variable is consulted; this shold be a colon-separated list of directories, like the Unix PATH variable. If that is not set then the default directory (installation-specific but often **/usr/local/ssl/certs**) is processed.

Unrelated to this PR but related to PA-3541, I believe we should consider renaming or symlinking the ca-bundle file. I will comment about that in Jira.

gimmyxd (approved):

PR #417, opened 2021-01-28
(maint) Update puppet to 7.3.0 for bolt-runtime

beechtom:
This updates the bolt-runtime project to include Puppet 7.3.0.

Part of puppetlabs/bolt#2547

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -1,6 +1,16 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
-  pkg.version "6.20.0"
-  pkg.md5sum 'b1a6f244663f04075bafb1d36eede31c'
+  # Projects may define a :rubygem_puppet_version setting, or we use 6.20.0 by default:
+  version = settings[:rubygem_puppet_version] || '6.20.0'
+  pkg.version version
+
+  case version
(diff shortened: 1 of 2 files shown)


configs/components/rubygem-puppet.rb
```diff
@@ -1,6 +1,16 @@
...
-  pkg.version "6.20.0"
-  pkg.md5sum 'b1a6f244663f04075bafb1d36eede31c'
+  # Projects may define a :rubygem_puppet_version setting, or we use 6.20.0 by default:
+  version = settings[:rubygem_puppet_version] || '6.20.0'
+  pkg.version version
+
+  case version
+  when '7.3.0'
+    pkg.md5sum '82857bbc5a75c3b6e49ce15c18959504'
+  when '6.20.0'
```
lucywyman, line 9: This is going to force everyone else to configure this - I think making `6.20` the default would be better.

beechtom: ^ It is the default

```
version = settings[:rubygem_puppet_version] || '6.20.0'
```

lucywyman: Oh weird...sorry, we don't usually do it that way. Didn't even look!

lucywyman (approved):

PR #420, opened 2021-02-02
Add debian-10-armhf build

stahnma:
I wanted Raspberry Pi agents to be modern. This change allows the
puppet-runtime to build on debian-10-armhf on a direct compile (not cross)
because modern arm systems aren't totally terrible in terms of speed.

### DIFF
--- a/configs/components/openssl-1.1.1.rb
+++ b/configs/components/openssl-1.1.1.rb
@@ -78,6 +78,8 @@ component 'openssl' do |pkg, settings, platform|
       target = 'linux-ppc64le'
     elsif platform.architecture =~ /64$/
       target = 'linux-x86_64'
+    elsif platform.architecture == 'armhf'
+      target = 'linux-armv4'
     end
   end
 
--- a/configs/projects/_shared-agent-settings.rb
+++ b/configs/projects/_shared-agent-settings.rb
@@ -99,7 +99,7 @@ platform_triple = "ppc64-redhat-linux" if platform.architecture == "ppc64"
 platform_triple = "ppc64le-redhat-linux" if platform.architecture == "ppc64le"
 platform_triple = "powerpc64le-suse-linux" if platform.architecture == "ppc64le" && platform.name =~ /^sles-/
 platform_triple = "powerpc64le-linux-gnu" if platform.architecture == "ppc64el"
-platform_triple = "arm-linux-gnueabihf" if platform.name == 'debian-8-armhf'
+platform_triple = "arm-linux-gnueabihf" if platform.architecture == "armhf"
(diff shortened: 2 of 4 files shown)


stahnma:
Please look over the commit on `runtime.sh` with extra care. I worked for a quite a while to try to come with a better solution there and it just didn't work. It looks to me like there's a combination of not using pl-build-tools, not using internal mirrors, and not cross compiling that caused a weird state, but I'm not positive on it. I solved the problem, but I might have created new ones, so this is the one thing in this entire patch set that could use some better eyes.

stahnma:
On the Azure failure, I don't think it's related to any change I made.

resources/files/runtime/runtime.sh
```diff
@@ -16,7 +16,7 @@ else
     LIBDIR="$1"
   else
     echo "$1 does not exist" >2
-    exit 1
+    exit 0
```
gimmyxd, line 19: I think that we bypassed this by not running this sh on certain platforms. I don't think this change should affect any of existing platforms

configs/components/openssl-1.1.1.rb
```diff
@@ -78,6 +78,9 @@
       target = 'linux-ppc64le'
     elsif platform.architecture =~ /64$/
       target = 'linux-x86_64'
+    # Not a cross-compile for debian-10-arm
+    elsif platform.name =~ /debian-10-arm/
```
GabrielNagy, line 82: I know this has been done in a similar way for Debian 8, but it'd be more future-proof to change this `if` to something like `platform.architecture =~ /armhf/`.

configs/projects/_shared-agent-settings.rb
```diff
@@ -99,7 +99,7 @@
 platform_triple = "ppc64le-redhat-linux" if platform.architecture == "ppc64le"
 platform_triple = "powerpc64le-suse-linux" if platform.architecture == "ppc64le" && platform.name =~ /^sles-/
 platform_triple = "powerpc64le-linux-gnu" if platform.architecture == "ppc64el"
-platform_triple = "arm-linux-gnueabihf" if platform.name == 'debian-8-armhf'
+platform_triple = "arm-linux-gnueabihf" if platform.name =~ /debian-[\d?\d]-armhf/
```
GabrielNagy, line 102: Same as my other comment, we can use `platform.architecture =~ "armhf"` as a condition.

GabrielNagy (commented):
Just some minor comments, but should be good to merge otherwise!

stahnma (commented):
Changes made and squashed into original arm commit.

GabrielNagy (approved):

PR #421, opened 2021-02-03
(maint) Update bolt-runtime to use --no-document

beechtom:
This updates the bolt-runtime to use the `--no-document` option
when installing gems instead of the defunct `--no-rdoc` and `--no-ri`
options.


lucywyman (approved):
[Looks like this was added](https://github.com/rubygems/rubygems/blob/master/CHANGELOG.md#200--2013-02-24) in Ruby / Rubygems 2.0.0, should be safe with 2.5 and 2.7.

PR #422, opened 2021-02-04
(maint) Add rubygem-scanf component to bolt-runtime

beechtom:
This adds the rubygem-scanf component to bolt-runtime.


lucywyman:
Built great for me!

PR #384, opened 2020-10-27
(packaging) Add bootsnap dependency to bolt-runtime

jpogran:
Needed for https://github.com/puppetlabs/bolt/pull/2291

PR #429, opened 2021-03-08
(maint) Remove public chocolate source

jpogran:
During testing https://github.com/puppetlabs/puppet-runtime/pull/384 the public chocolate repo source was added, but I did not remove it before merging. This removes the public repo to ensure the correct sources are used.


jpogran:
@gimmyxd I've put up a commit that re-removes the public chocolatey source. This should fix what I broke

gimmyxd (approved):

PR #430, opened 2021-03-17
(maint) update git 2.31

jpogran:
Update Git version to 2.31


jpogran:
This was manually verified

GabrielNagy (approved):
Looks like only PDK uses the git component, so this looks sane to me :+1:
Feel free to merge whenever

PR #445, opened 2021-04-27
(maint) Bump Bolt and PE Bolt server dependencies

lucywyman:
This adds a new clause to allow projects to pull in Puppet 7.6.1 if they
have the version specified in the project file, and bumps dependencies
that are unique to Bolt and PE Bolt server. Shared dependency bumps are
in a separate PR for review by teams that need them.

Built and installed for ubuntu-20.04

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -4,8 +4,8 @@ component 'rubygem-puppet' do |pkg, settings, platform|
   pkg.version version
 
   case version
-  when '7.4.1'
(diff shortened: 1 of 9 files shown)


configs/components/rubygem-puppet.rb
```diff
@@ -4,6 +4,8 @@
   pkg.version version
 
   case version
+  when '7.6.1'
```
donoghuc, line 7: I think bolt is the only project that cares about this? We can probably just have a single 7 series version.

donoghuc (approved):

PR #463, opened 2021-06-24
(maint) Update Bolt-specific gem dependencies

lucywyman:
This updates gem dependency versions that are only used in Bolt
projects. Specifically this is to pull in orchestrator_client-ruby to
pick up a fix for the `URI.escape` warning.


lucywyman:
Successfully built, installed, and ran for Ubuntu 20.04 and Windows 2012r2

donoghuc (approved):

beechtom (approved):

PR #464, opened 2021-06-25
(packaging) bump resource-api to latest version

binford2k:
This version adds the custom `in_sync?` functionality, which allows
modern DSC resource types to work idempotently even when the underlying
DSC resource is flawed. Other types will likely start to use this soon
too.


binford2k:
This is already addressed in #463

PR #465, opened 2021-06-25
(maint) Remove container builds / tests

Iristyle:
- At one point it was thought that a shared based puppet-runtime
   container might be useful.

   That plan never materialized.

   Removing this now given it's tech debt and not used anywhere


Iristyle:
Going to merge... if there are any objections later, we can revert. But this container isn't being used for anything.

PR #467, opened 2021-07-01
(PE-31686) Ship puppet 6.23.0 with bolt/ace server

donoghuc:
Take up API changes for ace server to use v4 catalog endpoint more efficently.

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -1,13 +1,13 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
   # Projects may define a :rubygem_puppet_version setting, or we use 6.20.0 by default:
-  version = settings[:rubygem_puppet_version] || '6.20.0'
+  version = settings[:rubygem_puppet_version] || '6.23.0'
   pkg.version version
 
   case version
   when '7.7.0'
     pkg.md5sum '81cfcab6f6998bf7cfbcf0c91a72a469'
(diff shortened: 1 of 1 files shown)


steveax (approved):
👍

configs/components/rubygem-puppet.rb
```diff
@@ -1,13 +1,13 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
   # Projects may define a :rubygem_puppet_version setting, or we use 6.20.0 by default:
-  version = settings[:rubygem_puppet_version] || '6.20.0'
```
ekinanp, line 3: Hmm, any reason why we're not calculating this from `rubygem_puppet_version` ?

donoghuc: I figured we would do that work when we pull in the puppet7 in our components ticket. same as the gemspec change in the ace repo.

donoghuc: oh, hmm seems we need a ticket for that https://tickets.puppetlabs.com/browse/PE-31684

donoghuc: oh, nevermind, found it https://tickets.puppetlabs.com/browse/PE-31696

donoghuc: as part of that ticket we will move to setting :rubygem_puppet_version in the individual bolt-runtime project definitions.

PR #470, opened 2021-07-15
(PA-3792) Remove Fedora 30 x86_64 from puppet-runtime

BobosilaVictor:


gimmyxd:
There are some places where we do version checks for Fedora, we can get rid of those:
- https://github.com/puppetlabs/puppet-runtime/blob/14eaa1ff3ea384c30436e96ecd702b73d1a3f920/configs/components/runtime-pdk.rb#L36 (TBD with pdk team)
- https://github.com/puppetlabs/puppet-runtime/blob/c56420a650cfd5c3188fdbbdc6781ce886a193b0/configs/components/ruby-2.4.10.rb#L207 (TBD with pdk team)
- https://github.com/puppetlabs/puppet-runtime/blob/725f5bbba1c30aeffd86fd5cb49f91fcfeb88992/configs/components/ruby-2.7.3.rb#L93

joshcooper:
Same comment as https://github.com/puppetlabs/puppet-runtime/pull/472#issuecomment-987454008

lucywyman (approved):

joshcooper:
@da-ar gave :+1: in slack

PR #471, opened 2021-07-15
(maint) patch sys-filesystem gem

ciprianbadescu:
The commit https://github.com/djberg96/sys-filesystem/commit/4024d9569cc11841f605bb9f5e74e42c32f2082e causes crash on 32 bits Linux, we should use statvfs64 only on 64 bits Linux

Change also proposed upstream: https://github.com/djberg96/sys-filesystem/pull/56


ciprianbadescu:
jenkins please test agent-runtime-main with puppet-agent#main on all

luchihoratiu (approved):

PR #473, opened 2021-07-15
(maint) Pin FFI to 1.13.1 for the Bolt runtime

lucywyman:
This adds a new optional project setting `rubygem_ffi_version` that can
be set to '1.13.1' to continue using FFI 1.13.1 for projects where
upgrading to 15 breaks. We're expecting to be able to unpin the gem in
Bolt's gemspec and revert this change, but need a working build while
that's tested.

### DIFF
--- a/configs/components/rubygem-ffi.rb
+++ b/configs/components/rubygem-ffi.rb
@@ -1,5 +1,6 @@
 component "rubygem-ffi" do |pkg, settings, platform|
-  if platform.is_cross_compiled? && (platform.is_linux? || platform.is_solaris?)
+  if (platform.is_cross_compiled? && (platform.is_linux? || platform.is_solaris?)) ||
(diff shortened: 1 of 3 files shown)


donoghuc:
can you set this in bolt-server runtimes too please?

lucywyman:
Built, installed, and ran Bolt successfully for:
- [x] Ubuntu 20.04
- [ ] Windows 2012r2
- [ ] OSX 10.15

donoghuc (approved):
This looks like it will work. There is certainly quite a bit of complexity here.

lucywyman:
Ook should be gtg

donoghuc:
uhh @lucywyman i thought we were able to unpin this?

lucywyman:
Yeah I should have left a comment - I wanted to leave this open until we released Bolt just in case something went wrong, but we don't need this change. I'll revert.

PR #480, opened 2021-08-12
(PA-3883) Add GitHub Actions workflow that shows component diff

luchihoratiu:
This currently runs only with `puppet-runtime` as project but with all available platforms (retrieved from `configs/platforms`). Comment example can be seen [here](https://github.com/luchihoratiu/puppet-runtime/pull/2#issuecomment-904591632) with changes done in a3c136d201cb60654e2fbdcf258dcea7a85867dc.

### DIFF
--- a/.github/workflows/comment_on_pr.yaml
+++ b/.github/workflows/comment_on_pr.yaml
@@ -1,0 +1,79 @@
+---
+name: Comment output on PR
+
+on:
+  workflow_run:
+    workflows: ["Vanagon Component Diff"]
+    types:
+      - completed
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: 'Download artifacts'
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var artifacts = await github.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: ${{github.event.workflow_run.id }},
+            });
+            var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "artifacts"
+            })[0];
+            var download = await github.actions.downloadArtifact({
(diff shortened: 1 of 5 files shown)


tasks/vanagon_component_diff.rake
```diff
@@ -0,0 +1,164 @@
+# frozen_string_literal: true
```
gimmyxd, line 1: Would be nice to have this whole rake task in Vanagon, and then use it wherever we need.
It will be difficult to manage this functionality in multiple Vanagon projects.

luchihoratiu: Agreed, this was just to get a poc working.

luchihoratiu:
test

.github/workflows/comment_on_pr.yaml
```diff
@@ -0,0 +1,79 @@
...
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            });
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/artifacts.zip', Buffer.from(download.data));
+            fs.writeFileSync('./artifactid', matchArtifact.id)
+
+      - name: 'Unzip artifacts'
```
gimmyxd, line 37: Can we use https://github.com/actions/download-artifact ?

configs/components/curl.rb
```diff
@@ -45,7 +45,6 @@
 
   pkg.configure do
     ["CPPFLAGS='#{settings[:cppflags]}' \
-      LDFLAGS='#{settings[:ldflags]}' \
```
gimmyxd, line 48: I think this is not intended

.github/workflows/comment_on_pr.yaml
luchihoratiu: It seems that this action is useful for using in the same workflow and doesn't have a parameter to change workflow id from where to retrieve artifacts.

gimmyxd (approved):

PR #484, opened 2021-08-27
(maint) Update Bolt dependency gems

lucywyman:
Updates gem versions to their latest resolved versions for dependencies
of Bolt and pe-bolt-server.

Verified this builds + runs for both Bolt and pe-bolt-server packages.

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -1,13 +1,14 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
-  # Projects may define a :rubygem_puppet_version setting, or we use 6.20.0 by default:
-  version = settings[:rubygem_puppet_version] || '6.23.0'
+  # Projects may define a :rubygem_puppet_version setting, or we use this
+  # version by default
+  version = settings[:rubygem_puppet_version] || '6.24.0'
   pkg.version version
 
   case version
-  when '7.7.0'
-    pkg.md5sum '81cfcab6f6998bf7cfbcf0c91a72a469'
-  when '6.23.0'
-    pkg.md5sum '10dbdce584a286bed95835420e2fba48'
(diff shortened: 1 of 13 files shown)


beechtom (approved):

configs/components/rubygem-puppet.rb
```diff
@@ -4,8 +4,8 @@
   pkg.version version
 
   case version
-  when '7.7.0'
-    pkg.md5sum '81cfcab6f6998bf7cfbcf0c91a72a469'
+  when '7.10.0'
+    pkg.md5sum '801e1945b1c483d1d5a4cb9b1caf7578'
   when '6.23.0'
```
donoghuc, line 9: It would probably be good to get on latest 6.x puppet for PE.

lucywyman: d-d-d-done

configs/components/rubygem-puppet.rb
```diff
@@ -4,10 +4,10 @@
   pkg.version version
 
   case version
-  when '7.7.0'
-    pkg.md5sum '81cfcab6f6998bf7cfbcf0c91a72a469'
-  when '6.23.0'
-    pkg.md5sum '10dbdce584a286bed95835420e2fba48'
+  when '7.10.0'
+    pkg.md5sum '801e1945b1c483d1d5a4cb9b1caf7578'
+  when '6.24.0'
```
beechtom, line 9: Default version needs to be updated on line 3.

donoghuc:
I'm sure these are all mirrored in artifactory right?

lucywyman:
Yeah, mirroring is how I get the md5 sums for exactly that reason (so I know they're mirrored). I also got a successful builds of puppet-bolt and pe-bolt-server using these changes in the runtime.

donoghuc (approved):

PR #489, opened 2021-09-02
(maint) Show platform and project settings once per platform [skip-ci]

GabrielNagy:


GabrielNagy:
This should also make the hashdiff process less painful since there is less data to diff.

BobosilaVictor (approved):

luchihoratiu (approved):

PR #492, opened 2021-09-14
(maint) Remove bootsnap/msgpack from bolt-runtime

beechtom:
This removes the bootsnap and msgpack rubygem components from the
bolt-runtime project, as they will no longer be used by Bolt.


lucywyman (approved):
Confirmed no other projects use these gems.

PR #493, opened 2021-09-15
(PA-3769) Use correct openssl for RedHat 8 FIPS

GabrielNagy:


configs/components/openssl-1.1.1-fips.rb
```diff
@@ -27,26 +27,36 @@
...
   pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-force-fips-mode.patch'
-  pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-post-rand.patch'
   pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-spec-file.patch'
   pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-remove-env-check.patch'
-  pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-edk2-build.patch'
   pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1l-ec-group-new-from-ecparameters.patch'
   pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1l-sm2-plaintext.patch'
 
+  if platform.name =~ /-7-/
+    pkg.add_source 'file://resources/patches/openssl/openssl-1.1.1-fips-post-rand.patch'
```
GabrielNagy, line 36: This patch removes the `random.h` include which seems to fail on redhat8fips, so avoid patching it there. The patch below just accounts for the changes in this patch so it's not needed anymore either.

```
crypto/rand/rand_unix.c: In function 'syscall_random':
crypto/rand/rand_unix.c:402:57: error: 'GRND_NONBLOCK' undeclared (first use in this function); did you mean 'O_NONBLOCK'?
     return syscall(SYS_getrandom, buf, buflen, nonblock?GRND_NONBLOCK:0);
                                                         ^~~~~~~~~~~~~
                                                         O_NONBLOCK
```

configs/projects/_shared-agent-settings.rb
```diff
@@ -176,6 +176,14 @@
   proj.setting(:ldflags, "-L#{proj.libdir} -Wl,-rpath=#{proj.libdir},-z,relro,-z,now")
 end
 
+if platform.name =~ /^redhatfips-/
+  proj.setting(:openssl_version, '1.1.1-fips')
```
GabrielNagy, line 180: Previously the `openssl_version` setting was `1.1.1` on FIPS, which caused it to be installed via yum due to the `pkg.build_requires` statements. The vanagon logic first attempts to search for an included component (such as `openssl-1.1.1.rb` which is not used for RedHat FIPS), then falls back to installing the dependency with `yum`.

configs/projects/_shared-pe-bolt-server.rb
```diff
@@ -18,7 +18,7 @@
   exit(1)
 end
 
-if platform.name =~ /^redhatfips-7-.*/
+if platform.name =~ /^redhatfips-.*/
```
GabrielNagy, line 21: @nmburgan @actowery @barriserloth do you know if these system_openssl changes are still needed? This logic seems to have been added for most projects in https://github.com/puppetlabs/puppet-runtime/pull/228 and https://github.com/puppetlabs/puppet-runtime/pull/214.

For more context, we switched to building openssl 1.1.1 on redhatfips-7 where the system openssl is 1.0.2, and on redhatfips-8 where the system openssl is also 1.1.1. I'm not sure what the situation was back in 2019 when these changes were added.

ciprianbadescu (approved):

PR #496, opened 2021-09-21
(PE-32928) Ship PE Bolt Server with pre-release Puppet gem

lucywyman:
This updates the PE Bolt server runtime projects to set a pre-release
version of the Puppet 6.x series gem, and additionally updates the
Puppet rubygem component definition to use Artifactory when pulling in
the pre-release version instead of rubygems, since the gem won't be
published there.

This is in order to pull in changes are loading scripts in modules in
Puppet that are required for testing work in Orchestrator around loading
scripts. For now we plan to pin to this gem version since it should have
all of the changes we need, and doesn't need to be updated with new
changes.

Once Puppet 6.25 is availabe these changes can be removed.

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -9,9 +9,14 @@ component 'rubygem-puppet' do |pkg, settings, platform|
     pkg.md5sum '801e1945b1c483d1d5a4cb9b1caf7578'
   when '6.24.0'
     pkg.md5sum 'af6bbabf8ba8b11184f3ff143d4217ac'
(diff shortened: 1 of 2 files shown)


lucywyman:
Got a build going. I also mirrored this gem, so it's available in `buildsources` as well. We can have that be the URL, but most I didn't want to also have to modify the mirror attribute.

donoghuc (approved):

ekinanp (approved):

GabrielNagy (approved):

lucywyman:
Verified the failing Orchestrator acceptance tests pass with this change :+1:

PR #512, opened 2022-01-24
(maint) Make boost and yaml-cpp optional

stahnma:
This commit alters the project file for agent-runtime to make boost and
yaml-cpp optional using the `NO_PXP_AGENT` env variable which was
already available and honored in the puppet-agent itself.

If `NO_PXP_AGENT` is defined, the project will skip boost and yaml-cpp
since facter does not need those anymore.

### DIFF
--- a/configs/projects/agent-runtime-main.rb
+++ b/configs/projects/agent-runtime-main.rb
@@ -54,6 +54,7 @@ project 'agent-runtime-main' do |proj|
     proj.component 'rubygem-sys-filesystem'
   end
 
-  proj.component 'boost'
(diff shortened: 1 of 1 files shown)


stahnma:
I have totally signed the CLA (and many other IP rights) for Puppet as I worked there for 8+ years.

binford2k:
@stahnma we updated it to match the Apache CLA to make it easier for contributors to get it approved by their legal teams and removed the trivial patch exemption. Annoying, but it means that everyone has to acknowledge it. 😐

joshcooper (approved):

PR #514, opened 2022-02-01
(PA-4187) Limits Windows gems to only build on Win

mhashizume:
Addresses a licensing issue with Ruby gems specific to Windows that use the Artistic 2.0 license by preventing them from building on non-Windows platforms.

### DIFF
--- a/configs/projects/bolt-runtime.rb
+++ b/configs/projects/bolt-runtime.rb
@@ -199,10 +199,15 @@ project 'bolt-runtime' do |proj|
   proj.component 'rubygem-yard'
 
   # Core Windows dependencies
-  proj.component 'rubygem-win32-dir'
-  proj.component 'rubygem-win32-process'
-  proj.component 'rubygem-win32-security'
-  proj.component 'rubygem-win32-service'
+
+  # Omits from non-Windows platforms to avoid licensing issues
+  if platform.is_windows?
+    proj.component 'rubygem-win32-dir'
+    proj.component 'rubygem-win32-process'
+    proj.component 'rubygem-win32-security'
+    proj.component 'rubygem-win32-service'
+  end
+  
   proj.component 'rubygem-windows_error'
   proj.component 'rubygem-winrm'
   proj.component 'rubygem-winrm-fs'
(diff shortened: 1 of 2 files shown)


configs/projects/bolt-runtime.rb
```diff
@@ -199,13 +199,15 @@
...
-  proj.component 'rubygem-winrm-fs'
+  if platform.is_windows?
+    proj.component 'rubygem-win32-dir'
+    proj.component 'rubygem-win32-process'
+    proj.component 'rubygem-win32-security'
+    proj.component 'rubygem-win32-service'
+    proj.component 'rubygem-windows_error'
+    proj.component 'rubygem-winrm'
+    proj.component 'rubygem-winrm-fs'
+  end
```
joshcooper, line 210: This is a little surprising, but we only want to omit `rubygem-win32-*` on non-Windows. But we do want `rubygem-winrm*` on all hosts so that Windows and non-Windows hosts can connect to and execute commands on remote Windows hosts. I think `rubygem-windows_error` is a dependency of `rubygem-ruby_smb` (defined earlier) need the same treatment, since bolt uses them to transfer files between `* <-> Windows`.

mhashizume: Thanks for the quick review and explanation, I've updated my commit to only omit the `rubygem-win32-*` gems.

joshcooper (approved):
LGTM. Might be good to get someone from bolt team to review?

donoghuc (commented):
This is fine for bolt. Why does installer project need the windows gems? I thought those were primary platform only? https://jenkins-platform.delivery.puppetlabs.net/view/puppet-runtime/view/pe-installer-runtime/job/platform_pe-installer-runtime_runtime-vanagon-packaging_2019.8.x/

I think you can just never ship the win32 gems for installer runtime.

cthorn42:
I'm 99% sure the windows gems are in the installer project due to copy and paste, just something overlooked.
Cas is correct that we should be good to just remove that from the pe-installer-runtime-2019.8.x. Actually looking at pe-installer-runtime-main.rb we went ahead and removed the win32 gems from that. So it is only in the 2019.8.x branch that those remain. 
We should clean this up.

PR #515, opened 2022-02-12
(maint) Add debian 11 armhf platform definition

stahnma:
This commit adds the Debian 11 definition for armhf (commonly found in
Raspberry Pi).


configs/platforms/debian-11-armhf.rb
```diff
@@ -0,0 +1,34 @@
...
+    "gcc",
+    "swig",
+    "systemtap-sdt-dev",
+    "zlib1g-dev"
+  ]
+
+  plat.provision_with "export DEBIAN_FRONTEND=noninteractive && apt-get update -qq && apt-get install -qy --no-install-recommends #{packages.join(' ')}"
+
+  plat.vmpooler_template "debian-11-armhf"
+  plat.output_dir File.join("deb", plat.get_codename, "PC1")
```
joshcooper, line 33: I'm thinking `PC1` isn't needed as I don't see that in the arm64 version https://github.com/puppetlabs/puppet-runtime/blob/6a53e0224a2d3bd33e9d86152f37d2ec52d4f52f/configs/platforms/debian-11-amd64.rb#L33

PR #519, opened 2022-03-08
(PA-4323) Update nokogiri gem for main, drop support for 6.x

cthorn42:
This PR bumps nokogiri gem due to a security issue. The nokogiri minor
version bump drops support for Ruby 2.5. So we can't support nokogiri on
the 6.x puppet runtime.
Also the mini_portile2 needed to be bumped due to a dependacy from nokogiri.


cthorn42:
Tested this with `bundle exec vanagon build agent-runtime-main osx-11-x86_64 <osx-112-vm>`. The output showed nokogiri 1.13.3 downloaded and installed, and also the mini_portile2 2.8.0 installed.

joshcooper (approved):
I verified that `mini_portile2` and `nokogiri` are only used by the agent runtime (so we don't need to keep the old gem version/checksums).

PR #525, opened 2022-03-21
(maint) Add --use-system-powershell for Wix

nmburgan:
For some reason, the Chocolatey machinery seems to think Wix is failing to install, which in fact it is installing fine. The use of the --use-system-powershell flag bypasses whatever error is happening in the Chocolatey machinery and allows the build to continue.

### DIFF
--- a/configs/platforms/windows-2012r2-x64.rb
+++ b/configs/platforms/windows-2012r2-x64.rb
@@ -26,7 +26,7 @@ platform "windows-2012r2-x64" do |plat|
     "pl-zlib-#{self._platform.architecture}",
     "mingw-w64 -version 5.2.0 -debug",
     "mingw",
-    "Wix310 -version 3.10.2 -debug -x86"
+    "Wix310 -version 3.10.2 -debug -x86 --use-system-powershell"
   ]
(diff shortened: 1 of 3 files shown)


steveax (approved):
👍

nicklewis:
The phrase "for some reason" makes me nervous... With neither an explanation of what the problem is _or_ how this fixes it (one or the other is a good start), this seems like a somewhat risky change.

nmburgan:
Yeah, I don't know why it works or why it fails in the first place.  On the other hand, using an ancient version of WiX is bad too, in addition to a MinGW-w64 version that seems not to exist on the official site and is also ancient.  I think we need to overhaul our build on Windows platforms a bit.

nmburgan:
In the build, it says "wix310 has been installed" then has a failure calling Write-ChocolateyFailure, which seems like it could be a problem with the chocolateyInstall.ps1 script, or perhaps the wix installer is exiting in a way it doesn't expect.  Why this is happening now, I don't know.  Maybe chocolatey changes?

cthorn42 (approved):

PR #527, opened 2022-03-23
(PE-33126) Fix for building client-tools-runtime on redhatfips-8-x86_64

nmburgan:
The openssl-1.1.1 package needed as a build dep for curl was conflicting with krb5-devel. This specifically uses openssl-1.1.1k on redhatfips8, which plays nicely with krb5-devel.


nmburgan:
Tested locally, this works for redhatfips8.

cthorn42 (approved):

PR #528, opened 2022-04-07
(PE-33733) Include gems in installer-runtime that are present in agent-runtime

nmburgan:
The pe-installer-runtime needs to include the same gems that agent-runtime includes, since we are running plans via the Bolt provided by pe-installer, and if one of those plans depends on a gem vendored by Puppet, it should be able to rely on it being there as well. Additionally, this removes code for handling Windows and OSX builds for pe-installer-runtime, since we don't build for those platforms, and modifies the component pulling in pl-build-tools so specifically apply to platforms that need it, rather than everything except el8.

In the future, we can probably do away with pe-installer-runtime entirely and use the plain puppet-runtime. But for now, we'll continue to use it to minimize the number of changes happening at once.


jonathannewman (approved):
LGTM! Runtime?

PR #529, opened 2022-04-12
(PA-4364) Bump puppet-runtime's Ruby to 2.7.6

cthorn42:


cthorn42:
Tested this with, `bundle exec vanagon build agent-runtime-main el-7-x86_64 <centos 7 vm name>`, and built the agent runtime with a ruby of 2.7.6.

nmburgan (approved):

PR #531, opened 2022-04-12
(PA-4364) Patch Ruby 2.5.9 for buffer overrun vuln

cthorn42:


joshcooper (approved):
Lgtm might be good to build agent-runtime-6.x on at least one host

cthorn42:
Tested this with an el-7-x86_64, was able to build the 6.x agent-runtime

PR #532, opened 2022-04-14
(PE-33733) Include gems in installer-runtime that are present in agent-runtime, take 2

nmburgan:
Take 2 of https://github.com/puppetlabs/puppet-runtime/pull/528.  This removes selinux from the project definition, since it's already in the shared agent components, and adds the openssl fix that was also needed for client-tools-runtime.


nmburgan:
Tested building both pe-installer-runtime-2019.8.x and pe-installer-runtime-main for all primary platforms, and all build without error.

steveax (approved):
run!
LGTM!

PR #533, opened 2022-04-19
(PE-33792) Define net-ssh version before pulling in shared components

nmburgan:
Because defining the version of net-ssh happened after we pulled in the shared agent components, it ended up using the default of 4.2 rather than our desired 5.2. This moves the setting above where we pull in the shared components.


nmburgan:
Tested manually by building pe-installer-runtime and then building pe-installer pointing to this runtime, and installing it on a test node.  It makes the host key check issue go away.

jonathannewman:
LGTM! Gems?

PR #540, opened 2022-05-11
(PA-4440) Build java cert store whenever java is available

justinstoller:
Previously, we used a long regex to determine which agent platforms
we would build a cert store containing external ca roots for. However,
the allowlist regex is rarely updated and is significantly out of date.
To prevent the list becoming out of date this patch updates us to
conditionally build the keystore when java is present intead.

### DIFF
--- a/configs/components/puppet-ca-bundle.rb
+++ b/configs/components/puppet-ca-bundle.rb
@@ -16,10 +16,14 @@ component "puppet-ca-bundle" do |pkg, settings, platform|
     "#{platform[:make]} install OPENSSL=#{openssl_cmd} USER=0 GROUP=0 DESTDIR=#{File.join(settings[:prefix], 'ssl')}"
   ]
 
-  # Include the keystore for the agent if the agent has a java
+  # Include the keystore for the agent if the agent has a java and is NOT FIPS
   if settings[:runtime_project] == 'agent'
-    if platform.name =~ /el-(6|7|8)|sles-(11|12)|debian-(9|10)|ubuntu-(14|15|16|18)/
-      install_commands << "#{platform[:make]} keystore OPENSSL=#{openssl_cmd} DESTDIR=#{File.join(settings[:prefix], 'ssl')}"
+    # To build the type of FIPS compliant keystore PE uses requires the BCFIPS
+    # library. It is assumed that FIPS users do not, in general, want their hosts
+    # to easily connect to the larger Internet and so it is not worth the effort
+    # to support FIPS keystores in the agent builds. 
+    if platform.name !~ /fips/
+      install_commands << "if [[ `which keytool` ]]; then #{platform[:make]} keystore OPENSSL=#{openssl_cmd} DESTDIR=#{File.join(settings[:prefix], 'ssl')}; fi"
     end
   end
 


configs/components/puppet-ca-bundle.rb
```diff
@@ -18,9 +18,7 @@
 
   # Include the keystore for the agent if the agent has a java
   if settings[:runtime_project] == 'agent'
-    if platform.name =~ /el-(6|7|8)|sles-(11|12)|debian-(9|10)|ubuntu-(14|15|16|18)/
-      install_commands << "#{platform[:make]} keystore OPENSSL=#{openssl_cmd} DESTDIR=#{File.join(settings[:prefix], 'ssl')}"
-    end
+    install_commands << "if [[ `which keytool` ]]; then #{platform[:make]} keystore OPENSSL=#{openssl_cmd} DESTDIR=#{File.join(settings[:prefix], 'ssl')}; fi"
```
joshcooper, line 21: Ah yeah makes sense. Will this work for cross compiled platforms? Will Windows do something silly if Java happens to be installed?

justinstoller: wrt cross compiled I don't think is an issue because the default format for many platforms (and what the server expects to read the keystore as) is JKS, which I believe is a cross platform Java standard.

However, it looks like starting in Java 9 the default format for creating a keystore was changes PKCS12, which won't work for Puppet Server. I've created this PR to correct that: https://github.com/puppetlabs/puppet-ca-bundle/pull/14

Also, the JKS format isn't FIPS compatible and a FIPS PE build will refuse to load this file. I know if you tell Java to load a JKS file but actually give it a PKCS12 file it simply say it found no certs (but not explode), I need to verify that happens on FIPS as well. To generate the keystore type we use in FIPS (BCFKS) we need the BouncyCastleFIPS library available. I think skipping FIPS (or shipping a file it won't read) as part of this is fine as talking to arbitrary hosts on the internet isn't usually what FIPS users want anyways (quite the opposite). FIPS users may regenerate this themselves using the BCFIPS jars we provide, and without extra work FIPS users should still be able to authenticate to internal hosts using the `ssl_trust_store` option.

justinstoller: Confirmed that windows-2012r2-x64 and windows-2019-x64 puppet-runtime-main builds. `puppet-agent` builds with windows-2012r2-x64 then installs correctly but does not have a keystore in the location I would expect it `C:\Program Files\Puppet Labs\Puppet\puppet\ssl`
Screenshot from 2022-05-12 14-42-03

joshcooper: > I can add back a conditional to skip this on FIPS builds?

Yeah explicitly omitting the file sounds better than creating the file in a format that BouncyCastleFIPS can't load. So even if a user regenerates the file in the correct format, it'd be overwritten when the puppet-agent package is updated.

justinstoller: updated with a fips exclusion and a healthy comment as to why (updated commit message too).
built and confirmed no puppet-cacerts file on redhatfips-7-x86_64 (aptly named `epic-correction.delivery.puppetlabs.net`).

joshcooper (approved):

PR #541, opened 2022-05-12
(PA-4489) Bump nokogiri

cthorn42:
For the second time this month there is a security patch for Nokogiri.


joshcooper (approved):
I like your branch name :smile:

PR #545, opened 2022-06-06
(maint) Update gem components for bolt, bolt-server

beechtom:


configs/projects/pe-bolt-server-runtime-main.rb
```diff
@@ -1,6 +1,6 @@
 project 'pe-bolt-server-runtime-main' do |proj|
   proj.setting(:pe_version, 'main')
-  proj.setting(:rubygem_puppet_version, '7.13.1')
+  proj.setting(:rubygem_puppet_version, '7.17.0')
```
donoghuc, line 3: Might as well get on latest 6.x puppet gem for 2019.8.x too https://github.com/puppetlabs/puppet-runtime/blob/306a41ea7d17ae4ea57fb97c68b75c9f2efed019/configs/projects/pe-bolt-server-runtime-2019.8.x.rb#L3

PR #546, opened 2022-06-06
(maint) Add rexml rubygem component to bolt, bolt-server

beechtom:
This adds the rexml rubygem component which is a runtime dependency of
gyoku-1.4.0


beechtom:
Successfully built pe-bolt-server-runtime-2019.8.x 👍

PR #547, opened 2022-06-07
(maint) Downgrade multipart-post for bolt, bolt-server

beechtom:
This downgrades the `multipart-post` gem component for bolt and
bolt-server. Both projects pin the `faraday` gem component to the 0.x
series, which uses deprecated namespaces in newer versions of the
`multipart-post` gem, causing warnings to bubble up to the console when
running Bolt.

image


donoghuc (approved):
😢

PR #553, opened 2022-06-21
(maint) Remove rexml, webrick gems from bolt, pe-bolt-server

beechtom:
This removes the `rubygem-rexml` and `rubygem-webrick` components
from bolt and pe-bolt-server.

The `rexml` and `webrick` gems are bundled gems in Ruby, so do
not need to be included in the runtime build.


nicklewis:
I believe this is also true for `webrick` https://stdgems.org/2.7.6/

joshcooper (approved):
fwiw, webrick is no longer included in Ruby 3, so we'll need to add that back (I think?) when bolt-pe-server moves to Ruby 3

joshcooper:
Unfortunately it looks like webrick is still transitively needed: 

```
14:49:32 cd ./ && \
14:49:32 /opt/puppetlabs/bolt/bin/gem install --no-document --local --bindir=/opt/puppetlabs/bolt/bin puppet-strings-2.9.0.gem
14:49:32 ERROR:  Could not find a valid gem 'webrick' (~> 1.7.0) (required by 'puppet-strings-2.9.0.gem' (>= 0)) in any repository
14:49:34 make: *** [Makefile:2650: rubygem-puppet-strings-install] Error 2
```

PR #562, opened 2022-07-27
(PA-4445) Adds support for Fedora 36 (x86-64)

mhashizume:


mhashizume:
I was able to build both agent-runtime-6.x and agent-runtime-main with these changes.

joshcooper:
Since fedora 36 was added to vanagon 0.28.0, should we update the minimum version in https://github.com/puppetlabs/puppet-runtime/blob/82b38f1ed8e550728a45702a2c7d4de5a24762d5/Gemfile#L14?

joshcooper (approved):

mhashizume:
> Not blocking. Since fedora 36 was added to vanagon 0.28.0, should we update the minimum version in
> 
> https://github.com/puppetlabs/puppet-runtime/blob/82b38f1ed8e550728a45702a2c7d4de5a24762d5/Gemfile#L14
> ? On the other hand perfection is the enemy of progress

An easy enough fix, thanks for catching that

PR #563, opened 2022-08-03
(CT-212) Cross-compile libicu for macOS arm64

joshcooper:
libicu's runConfigureICU script doesn't support cross-compiling, so pass the
necessary options directly to `./configure`. Note we have to compile libicu
twice, for the current build and intended host as described in[1].

[1] https://unicode-org.github.io/icu/userguide/icu4c/build.html

### DIFF
--- a/configs/components/libicu.rb
+++ b/configs/components/libicu.rb
@@ -67,11 +67,35 @@ component 'libicu' do |pkg, settings, platform|
   end
 
   pkg.build do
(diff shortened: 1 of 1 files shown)


mhashizume:
Josh built client-tools-runtime-2019.8.x successfully with this change, I was able to build client-tools-runtime-main as well here: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/825/

mhashizume (approved):

PR #564, opened 2022-08-09
(PE-34232) update bolt gem components

austb:


configs/components/rubygem-puppet.rb
```diff
@@ -5,8 +5,8 @@
   pkg.version version
 
   case version
-  when '7.17.0'
-    pkg.md5sum '6d3090fed500e5efd9192f6383c5029a'
+  when '7.18.0'
+    pkg.md5sum '16e019b09d29592cdf43dbea72e14e1c'
   when '6.27.0'
```
donoghuc, line 10: Should we also bump the 6 series for 2019.8.x?

austb: yep, probably, I pushed an update to 6.28.0

donoghuc (approved):

PR #568, opened 2022-08-15
(PA-4494) Update agent runtime's augeas to 1.13.0

cthorn42:
The location for augeas' tarball has changed. The previous versions are still hosted
where we expect, but going forward we'll need to grab them from the Augeas Github repo.
This adds a bit more complexity in the augeas component, once all projects are on 1.13.0
we should be able to simplfy things.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -4,11 +4,18 @@ component 'augeas' do |pkg, settings, platform|
   pkg.version version
 
   case version
+  # Starting with the 1.13.0 release where the tarball is staged looks to have changed, now on github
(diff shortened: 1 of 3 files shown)


mhashizume (approved):
LGTM. Do we want to include this change in 6.x too now that it's in maintenance mode?

cthorn42:
I think it is worth getting it into 6.x, the request wasn't specific to which version. And then there is the issue that 1.12.0 and earlier has a different download URL. For simplicities sake I think trying to get more runtimes off of 1.12.0 and earlier is for the better.

PR #571, opened 2022-08-23
(PA-4620) Install ruby@2.5 from our homebrew-puppet tap

joshcooper:
When building `agent-runtime-6.x` on macOS 11 & 12 M1, we now install ruby@2.5 from our `homebrew-puppet` tap.

Note these changes would affect pdk-runtime too, but we don't currently support pdk on M1 (as it still relies on ruby@2.4 which was disabled a long time ago).

~~Blocked on https://github.com/puppetlabs/homebrew-puppet/pull/357~~

### DIFF
--- a/configs/components/_base-ruby.rb
+++ b/configs/components/_base-ruby.rb
@@ -48,7 +48,12 @@ elsif platform.is_windows?
 elsif platform.is_macos?
   pkg.environment 'optflags', settings[:cflags]
   if platform.is_cross_compiled?
-    pkg.build_requires "ruby@#{ruby_version_y}"
+    # Pin to an older version of ruby@2.5. This can be removed once we're no longer cross-compiling
+    if ruby_version_y == "2.5"
+      pkg.build_requires "puppetlabs/puppet/ruby@2.5"
+    else
+      pkg.build_requires "ruby@#{ruby_version_y}"
+    end
     pkg.environment 'CC', 'clang -target arm64-apple-macos11' if platform.name =~ /osx-11/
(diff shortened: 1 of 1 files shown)


configs/components/_base-ruby.rb
```diff
@@ -48,7 +48,8 @@
 elsif platform.is_macos?
   pkg.environment 'optflags', settings[:cflags]
   if platform.is_cross_compiled?
-    pkg.build_requires "ruby@#{ruby_version_y}"
+    # Pin to an older version of ruby@2.5. This can be removed once we're no longer cross-compiling
+    pkg.build_requires "joshcooper/puppet/ruby@#{ruby_version_y}"
```
joshcooper, line 52: We only want to do this for projects that rely on `ruby@2.5` like `agent-runtime-6.x` but not `agent-runtime-main`.

joshcooper:
Sucessfully built these combinations so far:

```
 bundle exec vanagon build agent-runtime-6.x osx-12-arm64
 bundle exec vanagon build agent-runtime-6.x osx-11-arm64
```

joshcooper:
We don't currently build the `pdk-runtime` on macOS 11/12 M1 (see https://github.com/puppetlabs/ci-job-configs/blob/a145197f08fee0e08f0305cee0c514acf9fcb2e9/jenkii/platform/projects/pdk.yaml#L59-L60). Note the pdk-runtime will also try to install ruby@2.4, which doesn't work:

```
upper-spaghetti:homebrew root# su test -c '/usr/local/bin/brew install ruby@2.4'
Running `brew update --auto-update`...
Error: ruby@2.4 has been disabled because it is not supported upstream!
```

mhashizume (approved):

PR #575, opened 2022-09-01
(PA-4561) Replaces Travis CI with GitHub Actions

mhashizume:


mhashizume:
It looks like everything works with this new workflow. I'm a little uncertain about Ruby versioning with this; unlike other workflows we have I don't think it's as important to test against multiple Ruby versions, so I just set it to 2.7 (the version our new LTS uses).

Happy to make changes if that seems problematic.

joshcooper (approved):

PR #578, opened 2022-09-08
(PA-4494) Switch agent-runtime 6.x back to Augeas 1.12.0

cthorn42:
Ruby had some errors building with the readline changes in OSX on 6.x.


joshcooper (approved):

cthorn42:
@joshcooper I pushed up a change after you approved. The issue is I can't figure out how to install readline only on agent-runtime-main, so instead I'm only passing in the flag to use readline if it is the 1.13.0 augeas version.
Since there is changes to main I went and built a generic vanagon job with runtime main, here are the results:
https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/1646/
And here is the results for agent-runtime-6.x with my changes:
https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/1644/

joshcooper (approved):
> how to install readline only on agent-runtime-main

Yeah.. I'm not super clear on that either. I think it's possible to access the project from within the component and if it's `agent-runtime-main` then do `pkg.build_requires('readline')` but this looks good too. Thanks for testing all of those combinations!

PR #580, opened 2022-09-15
ensure the augeas package url for 1.13 is used

h0tw1r3:
augeas 1.13 fails to download when not using the mirror. project has not released a package to their website yet.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -40,7 +40,7 @@ component 'augeas' do |pkg, settings, platform|
     extra_config_flags = platform.name =~ /solaris-11|aix/ ? " --disable-dependency-tracking" : ""
   end
 
-  pkg.url "http://download.augeas.net/augeas-#{pkg.get_version}.tar.gz"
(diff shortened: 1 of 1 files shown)


joshcooper:
@h0tw1r3 thanks for your PR. We recently fixed a bug in vanagon that prevented the fallback from working (https://github.com/puppetlabs/vanagon/commit/dbd27217d94da3b5545d7425165047c30cbd68ca) We also bumped augeas to 1.14.1 and changed how the download link https://github.com/puppetlabs/puppet-runtime/blob/57dfe7f279dec07aff29cdc0ce13e919f9a0cd3a/configs/components/augeas.rb#L26-L28 Please try out the latest vanagon and puppet-runtime to see if you're issue is resolved. If so, please close this.

traylenator:
Trying with 202308160 with addition of latest augeas component I can confirm augeas source is downloaded okay.

This can probably be closed.

joshcooper:
Thanks @traylenator, I'll close this

PR #581, opened 2022-09-15
(PUP-11625) fix augeas support on osx big sur

h0tw1r3:
configure shipped with augeas does not link with libtool properly on big sur.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -37,6 +37,13 @@ component 'augeas' do |pkg, settings, platform|
       end
     end
 
+    if platform.is_macos?
+      pkg.build_requires 'readline'
+      pkg.build_requires 'autoconf'
+      pkg.build_requires 'automake'
+      pkg.build_requires 'libtool'
+    end
+
     extra_config_flags = platform.name =~ /solaris-11|aix/ ? " --disable-dependency-tracking" : ""
   end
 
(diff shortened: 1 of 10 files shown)


configs/components/augeas.rb
```diff
@@ -117,6 +123,9 @@
     pkg.environment "LDFLAGS", settings[:ldflags]
   end
 
+  # fix libtool linking on big sur
+  pkg.configure { ["/usr/local/opt/bin/autoreconf --force --install"] } if platform.is_macos?
```
cthorn42, line 127: When I attempted to build runtime with this PR I ran into an issue with the path here, I had to set it to `/usr/local/bin/autoreconf`.

cthorn42:
With the change to the path I was able to build agent-runtime-main with this, and used that runtime to build a puppet-agent 7. However when I installed that agent on an osx-12-x86_64 VM I am still seeing the same link error:
```
/opt/puppetlabs/puppet/bin/augtool --help
dyld[503]: Library not loaded: /usr/local/opt/readline/lib/libreadline.8.dylib
  Referenced from: /opt/puppetlabs/puppet/bin/augtool
  Reason: tried: '/usr/local/opt/readline/lib/libreadline.8.dylib' (no such file), '/usr/local/lib/libreadline.8.dylib' (no such file), '/usr/lib/libreadline.8.dylib' (no such file)
Abort trap: 6
```

cthorn42:
Noticed in your Jira ticket that you were using ruby to apply the Augeas change, I can confirm when applying the test_ssh manifest that it works with your change. So this PR seems pretty close we just need to figure out the binary linkage.

h0tw1r3:
@cthorn42 the readline link appears to be unrelated to this PR. My tests indicate augtool is broken in the latest packaged release. I suspect the problem goes back to 1274c9a390cdf18e6f5533686be5475d3934ffad.

h0tw1r3:
I was able fix the link by creating a readline component. Does that make sense to do?

cthorn42:
That does make a lot of sense, I was planning on trying this in the next month or so. Let me kick off some automation to see if agent-runtime will build for all supported platforms with this change.

cthorn42:
These changes look great. I tested this out to build a agent-runtime-main for osx-12-x86_64, and then used then built a puppet-agent with that runtime. Augeas is fixed now. 
For testing purposes I kicked off a Jenkins job that built runtime for main and all runtimes were able to build. I just kicked off the agent-runtime-6.x validation, once we confirm all 6.x agent-runtimes build I'll merge this PR.

cthorn42 (approved):
the agent-runtime.6.x job passed and can confirm that both agent runtimes build with this changes for all supported platforms. 
@h0tw1r3 thanks for the contribution!

cthorn42:
@h0tw1r3 Sorry something I should have mentioned before I approved this needs to be rebased. If you could do that and push up the changes I'll merge it ASAP.

PR #589, opened 2022-10-11
(PA-4715) Add Ruby 3.2 to puppet-runtime, validated with EL-7 & EL-8

cthorn42:
This PR adds ruby 3.2 as a component to puppet runtime. It also copies the agent-runtime-main into a new project that has ruby 3.2. Until we have a branch of puppet-agent to promote ruby 2.7.6 to we'll need to seperate out this.
There are three changes that are needed to get Ruby 3.2 to build:
  * Package libffi
  * Package libyaml
  * Patch ruby-shadow Starting with Ruby 3.2 libffi and libyaml are no longer packaged with Ruby, so we need to take care of that. Ruby-shadow is a gem dependcy that is way out of date and wont build with Ruby 3, there is a simple patch we apply that allow us to build it.

The scope of this PR is just to build against EL-7 and EL-8, there is further work to do to get all other supported platforms to build.

### DIFF
--- a/configs/components/ruby-shadow.rb
+++ b/configs/components/ruby-shadow.rb
@@ -41,6 +41,15 @@ component "ruby-shadow" do |pkg, settings, platform|
     end
   end
 
+  matchdata = platform.settings[:ruby_version].match /(\d+)\.(\d+)\.\d+/
+  ruby_major_version = matchdata[1].to_i
+  if ruby_major_version >= 3
+    base = "resources/patches/ruby_32"
+    # https://github.com/apalmblad/ruby-shadow/issues/26
+    # if ruby-shadow gets a 3 release this should be removed
+    pkg.apply_patch "#{base}/ruby-shadow-2.5.0-cflags.patch", strip: "0"
+  end
+
   pkg.build do
     ["#{ruby} extconf.rb",
      "#{platform[:make]} -e -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
--- a/configs/projects/_shared-agent-components.rb
+++ b/configs/projects/_shared-agent-components.rb
@@ -13,6 +13,14 @@ end
 # Common components required by all agent branches
 proj.component 'runtime-agent'
 
+matchdata = platform.settings[:ruby_version].match /(\d+)\.\d+\.\d+/
+ruby_major_version = matchdata[1].to_i
+# Ruby 3.2 does not package these two libraries so we need to add them as a component
+if ruby_major_version >= 3
+  proj.component 'libffi'
+  proj.component 'libyaml'
+end
+
 if platform.name =~ /^redhatfips-.*/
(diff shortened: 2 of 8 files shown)


configs/projects/_shared-agent-components.rb
```diff
@@ -13,6 +13,14 @@
 # Common components required by all agent branches
 proj.component 'runtime-agent'
 
+matchdata = platform.settings[:ruby_version].match /(\d+)\.\d+\.\d+/
+ruby_major_version = matchdata[1].to_i
+# Ruby 3.2 does not package these two libraries so we need to add them as a component
+if ruby_major_version >= 3
+  proj.component 'libffi'
+  proj.component 'libyaml'
```
cthorn42, line 21: I really wanted to get this added in the ruby-3.2 component, but I couldn't make it work, so instead added it here with some logic gated to only ruby 3+

configs/components/ruby-shadow.rb
```diff
@@ -41,6 +41,15 @@
     end
   end
 
+  matchdata = platform.settings[:ruby_version].match /(\d+)\.(\d+)\.\d+/
+  ruby_major_version = matchdata[1].to_i
+  if ruby_major_version >= 3
+    base = "resources/patches/ruby_32"
+    # https://github.com/apalmblad/ruby-shadow/issues/26
+    # if ruby-shadow gets a 3 release this should be removed
+    pkg.apply_patch "#{base}/ruby-shadow-2.5.0-cflags.patch", strip: "0"
```
cthorn42, line 50: This part was tricky, usually when we apply_patch it defaults to strip: "1", which is non-relative pathing, but in this case vanagon has change directory into the ruby-shadow directory and we want to apply relative, so strip needs to be set to 0, and apply_patch takes an option hash to allow us to set that.

cthorn42:
Eventually when we start to promote agent-runtime-7.x into 2021.8.x PE, and we want to cut over to ruby 3.2, the configs/projects/agent-runtime-main_ruby_32.rb will be renamed to configs/projects/agent-runtime-main.rb. I figure in the mean time we can land ruby_32 changes and test them out.

joshcooper:
Yeah, it's unfortunate our 7.x builds are squatting on the `main` name... In the near future, it'd be good to rename the `agent-runtime-main` project to `agent-runtime-7.x` and change CJC simultaneously:

```yaml
        - 'puppet-runtime-pipeline':
            <<: *main_agent_runtime_targets
            p_pipeline_qualifier: 'main'
            p_puppet_runtime_project_name: 'agent-runtime-7.x'
            p_promotion_target_branch: 'main'

```

cthorn42:
The rename already took place: https://github.com/puppetlabs/puppet-runtime/commit/78ad2321e31c0a10c8687d6cb09d63e690938b95 so at this point we just need the CJC change.

joshcooper (approved):

PR #590, opened 2022-10-19
(maint) Changes needed for AIX to compile Ruby 3.2.0-preview2

cthorn42:


configs/components/libffi.rb
```diff
@@ -29,7 +29,7 @@
   pkg.build_requires "runtime-#{settings[:runtime_project]}"
 
   pkg.configure do
-    ["./configure --prefix=#{settings[:prefix]} --sbindir=#{settings[:prefix]}/bin --libexecdir=#{settings[:prefix]}/lib/libffi --disable-multi-os-directory #{settings[:host]}"]
+    ["MAKE=\"gmake\" ./configure --prefix=#{settings[:prefix]} --sbindir=#{settings[:prefix]}/bin --libexecdir=#{settings[:prefix]}/lib/libffi --disable-multi-os-directory #{settings[:host]}"]
```
joshcooper, line 32: Suggest defining the MAKE environment variable earlier in:

```ruby
if platform.is_aix?
  pkg.environment "PATH", "/opt/pl-build-tools/bin:$(PATH)"
  pkg.environment "MAKE", platform[:make]
```

As we set it in https://github.com/puppetlabs/puppet-runtime/blob/master/configs/platforms/aix-7.1-ppc.rb#L4

joshcooper:
Looks like this has been overtaken by recent events.

PR #592, opened 2022-10-26
(PA-4717) Build Ruby 3.2 runtime for Windows

AriaXLi:
libffi and libyaml are unable to find mingw64 when configuring and building, so the path for mingw64 must be set before running those commands. Aditionally, Ruby includes esent.h for WCHAR, which is proivded by mingw-w64 but the mingw version used in puppet-runtime is too old to include that header. So, a patch is needed to include basetsd.h for WCHAR.

This PR is still a **work in progress**, there are still some errors preventing Ruby 3.2 runtime to build on Windows.

### DIFF
--- a/configs/components/libffi.rb
+++ b/configs/components/libffi.rb
@@ -28,12 +28,25 @@ component 'libffi' do |pkg, settings, platform|
 
   pkg.build_requires "runtime-#{settings[:runtime_project]}"
 
+  if platform.is_windows?
+    # In Windows, libffi is unable to find the C programs (like mingw) to configure and build so the path must be set first
+    c_tools_path = 'export PATH="/cygdrive/c/tools/mingw64/bin:$(PATH)"'
+  else
+    c_tools_path = ""
+  end
+
   pkg.configure do
-    ["./configure --prefix=#{settings[:prefix]} --sbindir=#{settings[:prefix]}/bin --libexecdir=#{settings[:prefix]}/lib/libffi --disable-multi-os-directory #{settings[:host]}"]
+    [
+      "#{c_tools_path}",
+      "./configure --prefix=#{settings[:prefix]} --sbindir=#{settings[:prefix]}/bin --libexecdir=#{settings[:prefix]}/lib/libffi --disable-multi-os-directory #{settings[:host]}"
+    ]
   end
 
   pkg.build do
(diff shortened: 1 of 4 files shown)


AriaXLi:
When running `./miniruby.exe,` it errors with `mv: cannot stat 'C:/ProgramFiles64Folder/PuppetLabs/Puppet/puppet/bin/erb.cmd': No such file or directory`

configs/components/ruby-3.2.0.rb
```diff
@@ -69,6 +69,10 @@
  #   pkg.apply_patch "#{base}/win32_long_paths_support.patch"
  #   pkg.apply_patch "#{base}/ruby-faster-load_27.patch"
  #   pkg.apply_patch "#{base}/windows_configure.patch"
+    # In Ruby 3.0, they added the esent.h for WCHAR to be included which is provided by 
+    # mingw-w64, but puppet-runtime uses an old version of mingw that's missing esent.h.
+    # To work around this, we will include basetsd.h. See patch for more information.
+    pkg.apply_patch "#{base}/win32-only-include-base-windows-types.patch"
```
joshcooper, line 75: We can drop the patch as it was merged upstream https://github.com/ruby/ruby/blob/v3_2_0/win32/dir.h#L4

configs/components/libffi.rb
```diff
@@ -28,12 +28,25 @@
 
   pkg.build_requires "runtime-#{settings[:runtime_project]}"
 
+  if platform.is_windows?
+    # In Windows, libffi is unable to find the C programs (like mingw) to configure and build so the path must be set first
+    c_tools_path = 'export PATH="/cygdrive/c/tools/mingw64/bin:$(PATH)"'
+  else
+    c_tools_path = ""
+  end
+
```
joshcooper, line 37: I think you'll want to override the PATH above on line 24 like:

```ruby
elsif platform.is_windows?
  pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
else
  ..
```
That way the `PATH` override will apply to the configure, build and install steps in this component.

mhashizume:
Closing this PR in favor of https://github.com/puppetlabs/puppet-runtime/pull/632

PR #593, opened 2022-10-26
(PA-4767) Update Nokogiri for CVE-2022-2309

mhashizume:
Nokogiri includes libxml2, older versions of which are vulnerable to CVE-2022-2309 NULL pointer derefence.

This commit updates Nokogiri to 1.13.9, which includes a patched version of libxml2.


mhashizume:
I was able to successfully build agent-runtime-main on all supported macOS platforms with the Vangon generic builder

joshcooper (approved):

PR #595, opened 2022-11-02
(maint) Update the runtime for Solaris now that we're using version 11.4

cthorn42:
Two changes so far that get things passing for 6.x, but still issues for main:

- Curl failed to build with an error around not finding nghttp2, by default it used to try to not build with that. Add a configure option flag to be explict about don't build with nghttp2.
- libxsxlt failed with: ```ld: fatal: option --version-script requires option -z gnu-version-script-compat to be specified```. Need to pass in some load flags to get that to pass.


cthorn42:
We went back to using solaris 11.1 images to build, this PR isn't needed.

PR #596, opened 2022-11-04
(PE-34844) Stop compiling bolt/ace-server gems against system openssl

donoghuc:
Previously we were using the `systemssl` flag when building shared bolt/ace server runtime artifacts. This should no longer be necessary as the vendored openssl versions should be compatible.


donoghuc:
@nmburgan Following up on https://puppet.slack.com/archives/CGJ0GTF4Y/p1666800364406759 Once merged i'll keep an eye that we dont regress on https://tickets.puppetlabs.com/browse/PE-27107 (which is why we originally started doing this).

justinstoller (approved):

PR #605, opened 2022-11-29
(PA-4805) Bump Ruby to 2.7.7 for runtime-main

cthorn42:
Add a patch for Solaris to get a `-std=c99` flag added


joshcooper (approved):
For posterity, here's where ruby adds `c99` https://github.com/ruby/ruby/blob/168ec2b1e5ad0e4688e963d9de019557c78feed9/tool/transform_mjit_header.rb#L241

PR #606, opened 2022-11-29
(MAINT) pdk-runtime vnext

chelnak:
PDKs runtime hasn't been touched in a long time. This PR is preparation for the next version of PDK.

We bump the default ruby used by pdk to 2.7.6 and attempt to align the runtime with others in this repository.

Additionally this PR contains two new projects; `pdk-runtime-user` and `pdk-runtime-system`. These exist for windows packaging purposes.

They simply add a higher level base_dir setting that will direct the sub project to build in either `LocalAppDataFolder` or `ProgramFiles64Folder`.

Consuming projects can then use these base directories and the settings derived from the accordingly accordingly.

Currently my outstanding question is how do we build the additional projects in to the current packaging process without over complicating things?

### DIFF
--- a/configs/projects/pdk-runtime.rb
+++ b/configs/projects/pdk-runtime.rb
@@ -17,11 +17,12 @@ project 'pdk-runtime' do |proj|
 
   # These flags are applied in addition to the defaults in configs/component/openssl.rb.
   proj.setting(:openssl_extra_configure_flags, [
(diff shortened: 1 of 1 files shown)


chelnak:
Changes from https://github.com/puppetlabs/puppet-runtime/pull/605 should ideally be added to this pr

PR #610, opened 2022-12-16
(PA-4815) Updates fast_gettext gem to 2.2.0

mhashizume:
Puppet projects will soon be using Ruby 3.2, which is not compatible with versions of fast_gettext earlier than 2.1 because of the deprecated taint checking features.

This commit also removes the conditional for multiple versions of fast_gettext, which was originally added in 29db591e55448931dabcc4803385c10b999ee1b7 for puppet-agent 1.10.x, which has long been deprecated.

### DIFF
--- a/configs/components/rubygem-fast_gettext.rb
+++ b/configs/components/rubygem-fast_gettext.rb
@@ -1,14 +1,14 @@
 component "rubygem-fast_gettext" do |pkg, settings, platform|
-  version = settings[:rubygem_fast_gettext_version] || '1.1.2'
+  version = settings[:rubygem_fast_gettext_version] || '2.2.0'
   pkg.version version
 
   case version
-    when '1.1.0'
-      pkg.md5sum "fc0597bd4d84b749c579cc39c7ceda0f"
-    when '1.1.2'
-      pkg.md5sum "df5a2462d0e1d2e5d49bb233b841f4d6"
-    else
-      raise "rubygem-fast_gettext version #{version} has not been configured; Cannot continue."
+  when '2.2.0'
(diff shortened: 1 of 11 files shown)


mhashizume:
Currently building a few runtimes in the vanagon-generic-builder to test this.

joshcooper (approved):
[fast_gettext 2.2.0](https://rubygems.org/gems/fast_gettext/versions/2.2.0) supports Ruby >= 2.5, so I think it should be compatible with all of the runtimes we currently support (since we dropped agent-runtime-5.5.x and ruby 2.1.x) Also thankfully fast_gettext doesn't have any runtime dependencies. As long as pdk and bolt runtimes still work, I'm :+1:

donoghuc:
Yeah, was just going to ask about this. Looks like WRT bolt we would need to make sure that `puppet` and `r10k` will be able to sue this version. Not sure if those gems already have releases allowing the 2.x series.

mhashizume:
Converting this to a draft as I work out the dependencies.

mhashizume:
@donoghuc good call Cas. Josh updated both puppet and r10k's gemspecs and I pushed out a new r10k release earlier this week.

This PR only updates these components for Bolt and the agent runtime that uses Ruby 3.2, both of which build successfully.

All of the other runtimes (PDK and agents 6.x and 7.x) should not change as a result of this PR, but I verified that they build in the Vanagon generic builder regardless.

joshcooper:
@donoghuc if you have a minute, could you review?

donoghuc (approved):
This looks correct, but why did you switch to validating against sha instead of md5 hashes? I think all the other rubygems are verified against md5...

PR #611, opened 2023-01-07
(PA-4815) Pins gettext versions for runtime-main

mhashizume:
gettext and related components have been updated by default in preparation for Ruby 3.2, but agent-runtime-main is still running Ruby 2.7. This commit pins gettext, fast_gettex, and gettext-setup to older versions in the runtime.


joshcooper (approved):
I verified that these versions are contained in the respective `rubygem-*.rb` components

PR #612, opened 2023-01-09
(maint) Fix AIX GCC options

mhashizume:
Recently, the GCC package for AIX changed from linking to GCC 8 to GCC 10, breaking compilation for things like Ruby native extensions and C++ components like Boost.

This commit specifies GCC 8 to restore working behavior for building runtimes on AIX.


mhashizume:
I've tested this locally building agent-runtime-6.x on AIX 7.1 PPC

cthorn42 (approved):

PR #613, opened 2023-01-10
(PA-4820) Merge agent-runtime-main_ruby_32 into agent-runtime-main

joshcooper:
Bump agent-runtime-main project to ruby 3.2

Remove gettext pins so that we default to the more recent versions. Also add the
new erubi and prime gem dependencies that gettext requires.

Delete agent-runtime-main_ruby_32 project.

---

Until now we've been getting away with running the `configure` script in
the ruby tarball. However, the script in the ruby 3.2.0 tarball assumes gnu11.
This commit switches to 3.2.0 and runs `autogen.sh` prior to running `configure`
as documented in ruby[1].

[1] https://docs.ruby-lang.org/en/master/contributing/building_ruby_md.html

---

Update ruby-shadow patch to se RbConfig::CONFIG not CONFIG and combine the
changes in the cflags.patch into one hunk.


joshcooper:
Blocked on https://github.com/puppetlabs/ci-job-configs/pull/8681

joshcooper:
Successfully built agent-runtime-main in this PR on most linux platforms in https://jenkins-platform.delivery.puppetlabs.net/view/puppet-runtime/view/agent-runtime-main/job/platform_agent-runtime_runtime-vanagon-packaging_main/340/

cthorn42:
My hacky workaround should be removed as part of this PR: https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/_base-ruby.rb#L9-L10

joshcooper:
Thanks for catching that @cthorn42! I was only building ruby 3.2 preview 2. Marked as a draft while I rebuild.

cthorn42 (approved):
Looks good to me, with something this complex that touches so much I think it is best to merge early and ticket out whatever failures we see come out from this.

PR #616, opened 2023-01-11
(PA-4815) Fix order of rubygem-prime include for pe-bolt-server

jpartlow:
The component needs to be added after the eval of
_shared-pe-bolt-server.rb which adds settings the component method relies on.


jpartlow:
Manual run of `bundle exec vanagon build pe-bolt-server-runtime-2021.7.x el-7-x86_64` passed for me locally.

mhashizume (approved):

PR #619, opened 2023-01-17
(PA-4843) Apply shadow/shadow.c patch too

joshcooper:
This pulls in the patch submitted to upstream for both files.


joshcooper:
I built the runtime on el-7-x86_64 and verified it resolved the facter acceptance test described in PA-4843

joshcooper:
After merging this and rebuilding the runtime, we can remove the pending test in https://github.com/puppetlabs/facter/blob/main/acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb

mhashizume (approved):

PR #620, opened 2023-01-20
(PA-4844) Patch the autogenerated selinux C extension

joshcooper:
Calling into selinux from Ruby 3.2 generates a warning:

    selinux.so: warning: undefining the allocator of T_DATA class swig_runtime_data

This is because ruby 3.2 requires C extensions to "either use rb_define_alloc_func() to overwrite it or rb_undef_alloc_func() to delete it."[1] The selinux C extension selinuxswig_ruby_wrap.c is autogenerated by swig from the selinuxswig_i template.

swig 4.1.1 contains the fix for this[2] but we are using an older version on RHEL7. So patch the autogenerated C extension, which needs to happen after the code is generated and before it's compiled.

[1] https://github.com/ruby/ruby/blob/6963f8f743b42f9004a0879cd66c550f18987352/doc/extension.rdoc#label-Write+the+C+Code
[2] https://github.com/swig/swig/commit/962f0900018ecb10287b327d261e424fa5ccb755

### DIFF
--- a/configs/components/_base-ruby-selinux.rb
+++ b/configs/components/_base-ruby-selinux.rb
@@ -3,6 +3,8 @@
 # load it with instance_eval. See ruby-x.y-selinux.rb configs.
 #
 
+pkg.add_source("file://resources/patches/ruby-selinux/selinuxswig_ruby_wrap.patch")
+
 # These can be overridden by the including component.
 ruby_version ||= settings[:ruby_version]
 host_ruby ||= settings[:host_ruby]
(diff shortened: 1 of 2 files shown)


joshcooper:
I built this manually and verified it doesn't generate the error we were seeing in https://tickets.puppetlabs.com/browse/PA-4844

```
[root@asocial-reef ~]# openssl sha1 /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/3.2.0/x86_64-linux/selinux.so
SHA1(/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/3.2.0/x86_64-linux/selinux.so)= a3dee1c0b0914dad0d6ae4afa62d72d0fd757b7d
[root@asocial-reef ~]# /opt/puppetlabs/puppet/bin/puppet apply -e 'notice($facts["ruby"])'
Notice: Scope(Class[main]): {platform => x86_64-linux, sitedir => /opt/puppetlabs/puppet/lib/ruby/site_ruby/3.2.0, version => 3.2.0}
```

mhashizume (approved):

PR #623, opened 2023-01-23
(PA-4845) Use full path to rm on Windows

cthorn42:
Our Cygwin setup is a bit hacky, and the 'rm' command isn't found when vanagon is running it for some odd reason. This PR will use the full path if the platform is windows.

### DIFF
--- a/configs/components/rubygem-gettext.rb
+++ b/configs/components/rubygem-gettext.rb
@@ -16,10 +16,12 @@ component "rubygem-gettext" do |pkg, settings, platform|
   gem_home = settings[:puppet_gem_vendor_dir] || settings[:gem_home]
   pkg.environment "GEM_HOME", gem_home
 
+  remove_file_command = platform.is_windows? ? '/usr/bin/rm' : 'rm'
+
   case version
   when '3.4.3'
     install do
-      "rm -f #{gem_home}/gems/gettext-3.4.3/test/fixtures/gtk_builder_ui_definitions.ui~"
+      "#{remove_file_command} -f #{gem_home}/gems/gettext-3.4.3/test/fixtures/gtk_builder_ui_definitions.ui~"
     end
   end
 end


cthorn42:
Tested this to build bolt-runtime for both debian-11 and windows-2012. I'm not 100% sure if this is the correct fix, since it seems vanagon isn't using the correct ENV path on Windows. but this will patch around it.

joshcooper:
It looks like the [base-rubygem definition](https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/_base-rubygem.rb#L20-L23) modifies overrides the `PATH` environment variable on Windows, which is why `/usr/bin` isn't in the `PATH` as can be seen from this output:

```
14:21:09 export PATH="/cygdrive/c/tools/mingw64/bin:/cygdrive/c/ProgramFiles64Folder/PuppetLabs/Bolt/bin:/cygdrive/c/ProgramFiles64Folder/PuppetLabs/Bolt/bin:/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0" && \
14:21:09 export GEM_HOME="C:/ProgramFiles64Folder/PuppetLabs/Bolt/lib/ruby/gems/2.7.0" && \
14:21:09 cd ./ && \
14:21:09 C:/ProgramFiles64Folder/PuppetLabs/Bolt/bin/gem.bat install --no-document --local --bindir=C:/ProgramFiles64Folder/PuppetLabs/Bolt/bin gettext-3.4.3.gem && \
14:21:09 rm -f C:/ProgramFiles64Folder/PuppetLabs/Bolt/lib/ruby/gems/2.7.0/gems/gettext-3.4.3/test/fixtures/gtk_builder_ui_definitions.ui~
```

I'm thinking we should fix the base-rubygem component to append `:$PATH` in https://github.com/puppetlabs/puppet-runtime/blob/93e9f65df205014c22645afd1278ea48bed195be/configs/components/_base-rubygem.rb#L22

cthorn42:
Ahh, that is a simple fix, created https://github.com/puppetlabs/puppet-runtime/pull/624 and closing this out. Thanks Josh.

PR #631, opened 2023-02-06
(PA-4265) Remove c_rehash binary

cthorn42:
c_rehash binary has been replaced by an openssl subcommand 'rehash'. The binary is still packaged and that can lead to some confusion since we add that to puppet-agent, so this will just remove the c_rehash binary.

### DIFF
--- a/configs/components/openssl-3.0.rb
+++ b/configs/components/openssl-3.0.rb
@@ -196,6 +196,8 @@ component 'openssl' do |pkg, settings, platform|
   #   install_commands << "rm -f #{settings[:prefix]}/bin/{openssl,c_rehash}"
   # end
 
+  install_commands << "rm -f #{settings[:prefix]}/bin/c_rehash"
+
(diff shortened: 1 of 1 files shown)


configs/components/openssl-3.0.rb
```diff
@@ -196,6 +196,10 @@
   #   install_commands << "rm -f #{settings[:prefix]}/bin/{openssl,c_rehash}"
   # end
 
+  if settings[:runtime_project] =~ /agent/
+    install_commands << "rm -f #{settings[:prefix]}/bin/c_rehash"
+  end
+
```
joshcooper, line 202: Since `agent-runtime-main` is the only consumer of openssl 3, maybe remove it unconditionally? That way when client tools, pe installer, etc migrate to openssl 3, we don't reintroduce the security issue?

joshcooper (approved):

PR #632, opened 2023-02-06
(PA-4717) Ruby 3.2 for Windows x64 and x86

joshcooper:
* Adds gcc bin directory to PATH for libffi and libyaml
* Drops unneeded patches: `configure.ac`, `win32_long_paths_support`, `windows_ruby_2.5_fixup_generated_batch_files`
* Patch mkmf.rb to build native extensions using [mingw-w64](https://www.mingw-w64.org/) (instead of mingw)
* Fix how ffi gem was pinned on < 2.7
* Use latest ffi gem (1.15.5) and compile native extension `ffi_c`
* Install libssp-0.dll needed for gcc's stack protection
* Patch libffi 3.4.3 on Windows x86

Built x64 and x86 packages in https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/1922/

I created a separate ticket (PA-4913) to add this CI (along with pxp-agent-vanagon, etc)

### DIFF
--- a/configs/components/libffi.rb
+++ b/configs/components/libffi.rb
@@ -21,6 +21,10 @@ component 'libffi' do |pkg, settings, platform|
       pkg.environment 'CC', 'clang -target arm64-apple-macos11' if platform.name =~ /osx-11/
       pkg.environment 'CC', 'clang -target arm64-apple-macos12' if platform.name =~ /osx-12/
     end
+  elsif platform.is_windows?
+    pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
+    pkg.environment "LDFLAGS", settings[:ldflags]
+    pkg.environment "CFLAGS", settings[:cflags]
   else
     pkg.environment "LDFLAGS", settings[:ldflags]
     pkg.environment "CFLAGS", settings[:cflags]
--- a/configs/components/libffi.rb
+++ b/configs/components/libffi.rb
@@ -25,6 +25,10 @@ component 'libffi' do |pkg, settings, platform|
     pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
     pkg.environment "LDFLAGS", settings[:ldflags]
     pkg.environment "CFLAGS", settings[:cflags]
+
+    if platform.architecture == "x86"
+      pkg.apply_patch "resources/patches/libffi/revert_clang_32bit.patch"
+    end
   else
     pkg.environment "LDFLAGS", settings[:ldflags]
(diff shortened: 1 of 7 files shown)


configs/components/ruby-3.2.0.rb
```diff
@@ -149,6 +151,7 @@
         --enable-shared \
         --disable-install-doc \
         --disable-install-rdoc \
+        --enable-mkmf-verbose \
```
joshcooper, line 154: I'm not sure this is working as expected

configs/components/libffi.rb
```diff
@@ -21,6 +21,8 @@
       pkg.environment 'CC', 'clang -target arm64-apple-macos11' if platform.name =~ /osx-11/
       pkg.environment 'CC', 'clang -target arm64-apple-macos12' if platform.name =~ /osx-12/
     end
+  elsif platform.is_windows?
+    pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
   else
     pkg.environment "LDFLAGS", settings[:ldflags]
     pkg.environment "CFLAGS", settings[:cflags]
```
joshcooper, line 28: We've lost `LDFLAGS` and `CFLAGS` by adding the conditional

configs/components/ruby-3.2.0.rb
```diff
@@ -64,11 +64,9 @@
   end
 
   if platform.is_windows?
- #   pkg.apply_patch "#{base}/windows_ruby_2.5_fixup_generated_batch_files.patch"
- #   pkg.apply_patch "#{base}/windows_nocodepage_utf8_fallback_r2.5.patch"
- #   pkg.apply_patch "#{base}/win32_long_paths_support.patch"
+    pkg.apply_patch "#{base}/windows_mingw32_mkmf.patch"
+    pkg.apply_patch "#{base}/windows_nocodepage_utf8_fallback_r2.5.patch"
  #   pkg.apply_patch "#{base}/ruby-faster-load_27.patch"
```
joshcooper, line 69: Need to confirm if this patch is still needed in https://github.com/ruby/ruby/blob/88e0862267cec6764ab9888f7522a40f2996e2c7/load.c#L690-L691

joshcooper: I filed this as https://tickets.puppetlabs.com/browse/PA-4911 as it'll be easier to do perf testing once we have puppet-agent builds with ruby 3.2 in them.

configs/components/libffi.rb
joshcooper: fixed

mhashizume (approved):

PR #637, opened 2023-02-16
(CONT-550) Reject dh-autoreconf RHEL 9

GSPatton:
dh-autoreconf is rejected for RHEL 8 so this commit rejects it in RHEL 9 to fix failing pdk Jenkins jobs.

### DIFF
--- a/configs/components/git.rb
+++ b/configs/components/git.rb
@@ -48,7 +48,7 @@ component "git" do |pkg, settings, platform|
     ]
   end
 
-  if platform.name == 'el-8-x86_64'
+  if platform.name == 'el-8-x86_64' || platform.name == 'el-9-x86_64'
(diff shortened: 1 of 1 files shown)


chelnak (approved):

cthorn42:
I think some more work needs to be done to get pdk on el-9 to compile. You can test this out locally if you have access to VMPooler with this command: `bundle exec vanagon build pdk-runtime el-9-x86_64`
That will checkout a el-9-x86_64 VM and build the pdk-runtime on it.
Currently that is failing after your changes with:
```
Error: Error downloading packages:
  Cannot download Packages/expat-devel-2.2.10-4.el9.x86_64.rpm: All mirrors were tried
```

chelnak:
Superseded by #641

PR #520, opened 2022-03-10
(maint) Remove agent-runtime 5.5.x and ruby 2.4.10

joshcooper:
This is blocked on dropping puppet 5 in pdk some time later in 2022

PR #638, opened 2023-02-23
(CONT-660) Switch default ruby

chelnak:
This commit switches the PDKs default ruby to 2.7.7. 2.5.9 has been moved in to the additional rubies hash.

### DIFF
--- a/configs/projects/pdk-runtime.rb
+++ b/configs/projects/pdk-runtime.rb
@@ -52,12 +52,12 @@ project 'pdk-runtime' do |proj|
   proj.setting(:includedir, File.join(proj.prefix, "include"))
   proj.setting(:bindir, File.join(proj.prefix, "bin"))
 
-  proj.setting(:ruby_version, "2.5.9")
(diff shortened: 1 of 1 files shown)


jordanbreen28 (approved):

joshcooper (approved):
We dropped ruby 2.5 from agent runtimes in https://github.com/puppetlabs/puppet-runtime/pull/520, so is it time to drop it from pdk too? FWIW it's been EOL for nearly 2 years, but I imagine some people are still on Puppet 6...

Changes look fine. Were you able to build the runtime using the vanagon generic builder job against all supported platforms?

chelnak:
For now we have to keep 2.5 because PDK needs to support Puppet 6 for a while longer.

I'm hoping to drop it soon though 🙂

Package builds have only been done on a small subset for this particular set of changes but I wouldn't mind running a test across the whole target set.

I'll do that before we merge.

chelnak:
Vanagon looking good from build #1945

PR #641, opened 2023-02-27
(CONT-550) Prep PDK for RHEL9

GSPatton:
Prior to this change, puppet-runtime would fail to build due package installation failures.

Specifically, the expat-devel package failed to install because it did not exist on our mirror. 

This change makes use of a newly created repo with all of the packages needed to build puppet-runtime.

Additionally dh-autoreconf is also removed as a dependency as per RHEL8

### DIFF
--- a/configs/components/git.rb
+++ b/configs/components/git.rb
@@ -48,7 +48,7 @@ component "git" do |pkg, settings, platform|
     ]
   end
 
-  if platform.name == 'el-8-x86_64'
+  if platform.name == 'el-8-x86_64' || platform.name == 'el-9-x86_64'
     build_deps.reject! { |r| r == 'dh-autoreconf' }
   end
 
--- a/configs/platforms/el-9-x86_64.rb
+++ b/configs/platforms/el-9-x86_64.rb
@@ -15,4 +15,5 @@ platform "el-9-x86_64" do |plat|
     zlib-devel
   )
   plat.provision_with("dnf install -y --allowerasing  #{packages.join(' ')}")
+  plat.provision_with("sed -i 's/beta-x86_64\/baseos\/x86_64/base/' /etc/yum.repos.d/localmirror-baseos.repo; sed -i 's/beta-x86_64\/appstream\/x86_64/appstream/' /etc/yum.repos.d/localmirror-appstream.repo")
 end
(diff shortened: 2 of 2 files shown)


chelnak (approved):
I am assuming that we will remove this once the RHEL base image has been updated?

joshcooper:
Hmm this doesn't seem to work, I don't think the single quotes are preserved in the way you're hoping:

```
$ bx vanagon build agent-runtime-main el-9-x86_64 --preserve always
...
sed: -e expression #1, char 22: unknown option to `s'
sed: -e expression #1, char 25: unknown option to `s'
Remote ssh command (dnf install -y --allowerasing gcc gcc-c++ autoconf automake createrepo rsync cmake make rpm-libs rpm-build rpm-sign libtool libarchive && dnf install -y --allowerasing  java-1.8.0-openjdk-devel libsepol libsepol-devel libselinux-devel pkgconfig readline-devel rpmdevtools swig systemtap-sdt-devel yum-utils zlib-devel && sed -i 's/beta-x86_64/baseos/x86_64/base/' /etc/yum.repos.d/localmirror-baseos.repo; sed -i 's/beta-x86_64/appstream/x86_64/appstream/' /etc/yum.repos.d/localmirror-appstream.repo) failed on 'root@surprise-laugh.delivery.puppetlabs.net'.
```

Is there a ticket for updating the RHEL9 image? I think we added RHEL9 support to the agent during their beta and never updated the template after GA. Also wondering if it'd be quicker to just update the template?

chelnak:
Updating the image was something that @isaac-hammes was looking at for us. 

This was meant to be a stop gap but maybe it's worth waiting.

Thanks!

isaac-hammes (approved):

isaac-hammes:
This works for me.

chelnak:
@isaac-hammes Just tested along with 3d1778b0eaa82c16c782b53d0cc1fdfe0596b2b8 cherry-picked from #637.

Seems to be blasting through now.

chelnak:
@GSPatton Can you rebase and possibly consider cherry-picking 3d1778b0eaa82c16c782b53d0cc1fdfe0596b2b8?

joshcooper (approved):
I was able to build agent-runtime-main on el-9-x86_64

PR #645, opened 2023-03-03
(packaging) Experiment with building pe-bolt-server with ruby3/puppet8

donoghuc:
The bolt-server package contains the bolt code along with bolt library code that powers bolt-server, orchestrator, puppetserver and ace-server. Attempt to build a bolt-server that is able to run with the ruby 3.2 interpreter in puppet8.

TODO:
- Figure out how to get a puppet8 gem.
- Determine if el7 is the right primary platform OS to work with (no openssl3 yet)
- Firm up agent install used to build runtime artifact (native extensions are compiled with this version).

### DIFF
--- a/configs/components/rubygem-facter.rb
+++ b/configs/components/rubygem-facter.rb
@@ -1,6 +1,6 @@
 component 'rubygem-facter' do |pkg, settings, platform|
-  pkg.version '4.2.11'
-  pkg.md5sum '5a387c10f5917d88b46cacd5da2de6f1'
+  pkg.version '4.3.0'
+  pkg.md5sum 'aef369eddbda0a53790b07ceea30449e'
 
   instance_eval File.read('configs/components/_base-rubygem.rb')
 end
--- a/configs/components/runtime-pe-bolt-server.rb
+++ b/configs/components/runtime-pe-bolt-server.rb
@@ -11,7 +11,7 @@ component "runtime-pe-bolt-server" do |pkg, settings, platform|
   artifactory_url = 'https://artifactory.delivery.puppetlabs.net/artifactory'
 
   if platform.is_rpm?
-    platform.add_build_repository "#{artifactory_url}/rpm_enterprise__local/#{settings[:pe_version]}/repos/#{platform.name}/#{platform.name}.repo"
+    platform.add_build_repository "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__local/development/puppet-agent/0ca106596dfb1bb1509e0a7b1c09e6e6907bd6fe/puppet-agent-8.0.0.203.g0ca106596-1.el7.x86_64.rpm"
   end
 
   if platform.is_deb?
--- a/configs/components/rubygem-addressable.rb
+++ b/configs/components/rubygem-addressable.rb
@@ -1,6 +1,6 @@
 component 'rubygem-addressable' do |pkg, _settings, _platform|
-  pkg.version '2.8.0'
-  pkg.md5sum '9cd03fc88dd69b3c3491ce73de27f2f7'
+  pkg.version '2.8.1'
+  pkg.md5sum '64cb545a6ddcfc93c68fe53ac594c5f7'
 
   instance_eval File.read('configs/components/_base-rubygem.rb')
 end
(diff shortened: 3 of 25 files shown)


configs/components/runtime-pe-bolt-server.rb
```diff
@@ -11,7 +11,7 @@
   artifactory_url = 'https://artifactory.delivery.puppetlabs.net/artifactory'
 
   if platform.is_rpm?
-    platform.add_build_repository "#{artifactory_url}/rpm_enterprise__local/#{settings[:pe_version]}/repos/#{platform.name}/#{platform.name}.repo"
+    platform.add_build_repository "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__local/development/puppet-agent/0ca106596dfb1bb1509e0a7b1c09e6e6907bd6fe/el-7-x86_64/puppet-agent-7.23.0.203.g0ca106596-1.el7.x86_64.rpm"
```
justinstoller, line 14: Can we set this to the main 8.x builds now?

configs/components/rubygem-facter.rb
```diff
@@ -1,6 +1,13 @@
 component 'rubygem-facter' do |pkg, settings, platform|
-  pkg.version '4.2.11'
-  pkg.md5sum '5a387c10f5917d88b46cacd5da2de6f1'
+  pkg.environment "GEM_HOME", settings[:gem_home]
+  pkg.environment "GEM_PATH", settings[:gem_home]
+  pkg.url("https://nightlies.puppet.com/downloads/gems/facter-nightly/facter-latest.gem")
```
justinstoller, line 4: Ooohhh, so we'll need to wait until.... tomorrow and then point this at the versions we release.

configs/components/rubygem-orchestrator_client.rb
```diff
@@ -1,6 +1,16 @@
 component 'rubygem-orchestrator_client' do |pkg, settings, platform|
-  pkg.version '0.6.1'
-  pkg.md5sum 'a556e34451e9985a518230d6d79e90a3'
+  pkg.environment "GEM_HOME", settings[:gem_home]
+  pkg.environment "GEM_PATH", settings[:gem_home]
+  pkg.url("https://github.com/donoghuc/orchestrator_client-ruby.git")
+  pkg.ref("ruby-3.2")
```
justinstoller, line 5: Will you open the changes there in a PR to the upstream repo?

configs/components/rubygem-puppet.rb
```diff
@@ -1,17 +1,13 @@
 component 'rubygem-puppet' do |pkg, settings, platform|
-  # Projects may define a :rubygem_puppet_version setting, or we use this
-  # version by default
-  version = settings[:rubygem_puppet_version] || '6.28.0'
-  pkg.version version
+  pkg.environment "GEM_HOME", settings[:gem_home]
+  pkg.environment "GEM_PATH", settings[:gem_home]
+  pkg.url("https://nightlies.puppet.com/downloads/gems/puppet8-nightly/puppet-latest.gem")
```
justinstoller, line 4: Same as the facter comment, will wait till tomorrow and then needs updating.

configs/components/rubygem-rubyntlm.rb
```diff
@@ -1,6 +1,16 @@
 component 'rubygem-rubyntlm' do |pkg, settings, platform|
-  pkg.version '0.6.3'
-  pkg.md5sum 'e1f7477acf8a7d3effb2a3fb931aa84c'
+  pkg.environment "GEM_HOME", settings[:gem_home]
+  pkg.environment "GEM_PATH", settings[:gem_home]
+  pkg.url("https://github.com/larskanis/rubyntlm.git")
```
justinstoller, line 4: We definitely shouldn't do, IMO. Pulling in randos branches that could change out from underneath us. If there's not a released gem for this and we have to build from source I think we should fork this repo and build off of a SHA pointing to our own repo.

configs/components/_base-rubygem.rb
```diff
@@ -24,6 +24,7 @@
 # When cross-compiling, we can't use the rubygems we just built.
 # Instead we use the host gem installation and override GEM_HOME. Yay?
 pkg.environment "GEM_HOME", settings[:gem_home]
+pkg.environment "GEM_PATH", settings[:gem_home]
```
justinstoller, line 27: I think a note as to why this is now necessary would be valuable. Eg, are there certain gems that were moved out of the stdlib that now require it? Was it a temp thing for the puppet and facter latest gems? Or maybe to build a gem from source? Or are you actually fixing an existing bug?

configs/components/rubygem-rubyntlm.rb
donoghuc: For sure, this was just to get a build to test. We have options here...

PR #647, opened 2023-03-08
Puppet 8/Ruby 3 testing

nmburgan:


configs/components/_base-rubygem.rb
```diff
@@ -24,6 +24,7 @@
 # When cross-compiling, we can't use the rubygems we just built.
 # Instead we use the host gem installation and override GEM_HOME. Yay?
 pkg.environment "GEM_HOME", settings[:gem_home]
+pkg.environment "GEM_PATH", settings[:gem_home]
```
donoghuc, line 27: I forget why I hacked this in at one point.... Not sure if you just copied that over but i'm not actually convinced this is needed.

PR #656, opened 2023-04-07
(PA-5367) Revert ruby's 3.2 default to local encoding

tvpartytonight:
When ruby changed the `Encoding.default_external` with this PR at https://github.com/ruby/ruby/pull/2877, it broke puppet testing for Japanese Windows. With the default encoding set to UTF-8, both facter and puppet would error. This commit to puppet-runtime will modify the way puppet builds ruby and revert that commit, allowing powershell to use the locale instead of defaulting to utf-8.


tvpartytonight:
I have built a puppet-runtime package with this change and built a puppet-agent with that runtime package; it passes the acceptance criteria set in the JIRA ticket.

AriaXLi (approved):

joshcooper (approved):

PR #657, opened 2023-04-10
(maint) update nokogiri for ruby 3.2 support

tvpartytonight:
The previous version of 1.13.10 only supported < 3.2.dev. The older 1.13.10 version also had a restriction of > 3.1, but the newer 1.14.2 supports ruby 2.7 through 3.2.


tvpartytonight:
I kicked off a build for just osx-12-arm and it succeeded and is available on the internal build server.

PR #662, opened 2023-05-01
(PE-35088) Ship bolt-server gems compatable with jruby 9.4

donoghuc:
In order to take up jruby 9.4 in orchestrator, ensure the library code we ship loaded my jruby is compatable with jruby 9.4.


configs/components/_base-rubygem.rb
```diff
@@ -24,6 +24,7 @@
 # When cross-compiling, we can't use the rubygems we just built.
 # Instead we use the host gem installation and override GEM_HOME. Yay?
 pkg.environment "GEM_HOME", settings[:gem_home]
+pkg.environment "GEM_PATH", settings[:gem_home]
```
donoghuc, line 27: I can actually remove this. will update

configs/projects/pe-bolt-server-runtime-main.rb
```diff
@@ -1,11 +1,13 @@
...
+  proj.setting(:rubygem_puppet_version, '7.24.0')
+  proj.setting(:rubygem_net_ssh_version, '7.0.1')
   # We build bolt server with the ruby installed in the puppet-agent dep. For ruby 2.7 we need to use a --no-document flag
   # for gem installs instead of --no-ri --no-rdoc. This setting allows us to use this while we support both ruby 2.5 and 2.7
   # Once we are no longer using ruby 2.5 we can update.
   proj.setting(:no_doc, true)
 
   instance_eval File.read(File.join(File.dirname(__FILE__), '_shared-pe-bolt-server.rb'))
   proj.component 'rubygem-prime'
+  proj.component('rubygem-rexml')
```
jpartlow, line 12: Where did this come from? Or what's depending on it now, I guess.

jpartlow (approved):
:+1:

configs/projects/pe-bolt-server-runtime-main.rb
jpartlow: (It's there for puppet-agent 8 in the future.)

donoghuc: yep, thanks I should have noted that in a comment for now. This will be for the puppet 8, when we do the work to compile native extensions and package gems using the puppe8 ruby 3.2 interpreter this will need to be there. I should have waited until then to add it, but its not hurting anything currently.

PR #667, opened 2023-05-15
(PA-5430) Patch augeas C extension

AriaXLi:
This issue is very similar to PA-4844 (#620) except it affects augeas instead of selinux

When puppet is run with Ruby 3.2, this warning is generated:
`/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/augeas.rb:48: warning: undefining the allocator of T_DATA class Augeas`

This commit patches the autogenerated augeas C extension so there is no warning for augeas

### DIFF
--- a/configs/components/_base-ruby-augeas.rb
+++ b/configs/components/_base-ruby-augeas.rb
@@ -3,6 +3,8 @@
 # load it with instance_eval. See ruby-x.y-augeas.rb configs.
 #
 
+pkg.add_source("file://resources/patches/augeas/ruby-augeas-0.5.0-patch_c_extension.patch")
+
 # These can be overridden by the including component.
 ruby_version ||= settings[:ruby_version]
 host_ruby ||= settings[:host_ruby]
@@ -72,6 +74,9 @@ end
 
 pkg.build do
   build_commands = []
+  if ruby_version =~ /^3/
+    build_commands << "#{platform.patch} --strip=2 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../ruby-augeas-0.5.0-patch_c_extension.patch"
+  end
   build_commands << "#{ruby} ext/augeas/extconf.rb"
   build_commands << "#{platform[:make]} -e -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"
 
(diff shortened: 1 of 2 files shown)


configs/components/_base-ruby-augeas.rb
```diff
@@ -74,6 +76,9 @@
   build_commands = []
   build_commands << "#{ruby} ext/augeas/extconf.rb"
   build_commands << "#{platform[:make]} -e -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"
+  if ruby_version =~ /^3/
+    build_commands << "#{platform.patch} --strip=2 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../ruby-augeas-0.5.0-patch_c_extension.patch"
+  end
```
joshcooper, line 81: I think you'll want to patch before running `make`, but I'm not sure if patching needs to come before `ruby ext/augeas/extconf.rb`?

AriaXLi: Ah okay, I'll move that. What would be the best way to check if I should patch before ruby ext/augeas/extconf.rb too?

joshcooper: One option is to build the runtime using `--preserve always` and then ssh to the build host and run (I think):

```
$ /opt/puppetlabs/puppet/bin/ruby -e "require 'augeas'"
```

In theory that should be sufficient for the augeas ruby bindings to load the native library and trigger the warning.

That said, it looks we do all of the patching early in the `ruby-shadow` component (yet another rubygem with native extensions that we patch): https://github.com/puppetlabs/puppet-runtime/blob/b14bfccd97d13eead785d7aa0ca589de79a36623/configs/components/ruby-shadow.rb#L44-L57 So maybe follow that?

AriaXLi: Hmm `$ /opt/puppetlabs/puppet/bin/ruby -e "require 'augeas'"` doesn't trigger a warning on a host that doesn't have the patch

AriaXLi:
Without my changes the warning will pop up:
```
[root@morose-varnish ~]# /opt/puppetlabs/puppet/bin/ruby -e "require 'augeas'; Augeas.open('/')"
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/augeas.rb:48: warning: undefining the allocator of T_DATA class Augeas
```

With my changes the warning does not pop up:
```
[root@insipid-calumny ~]# /opt/puppetlabs/puppet/bin/ruby -e "require 'augeas'; Augeas.open('/')"
[root@insipid-calumny ~]#
```

AriaXLi:
Successfully built agent-runtime-main with build targets from [agent-runtime-main](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-runtime/view/agent-runtime-main/) on vanagon-generic-builder: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/1047/

joshcooper (approved):

PR #670, opened 2023-05-16
(PA-5427) Enable dynamicbase on OpenSSL3 and Windows FIPS

joshcooper:
FIPS builds of openssl 2.x on Windows required libeay32.dll to be loaded at a fixed address 0x63000000. If it wasn't then FIPS_check_incore_fingerprint would generate an error. The openssl 3 fips provider doesn't have that issue so reenable dynamic base (ASLR).

- [x] [agent-runtime-main#1e2148d0f6c3cf13a2db6e9c1e9cc36a6c4047c3](https://builds.delivery.puppetlabs.net/puppet-runtime/1e2148d0f6c3cf13a2db6e9c1e9cc36a6c4047c3/artifacts)
- [x] [pxp-agent-vanagon#7fc114663c6ce804e6e23ed26bee41d97ad51884](https://builds.delivery.puppetlabs.net/pxp-agent/7fc114663c6ce804e6e23ed26bee41d97ad51884/artifacts)
- [x] [puppet-agent](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1117/)

### DIFF
--- a/configs/components/openssl-3.0.rb
+++ b/configs/components/openssl-3.0.rb
@@ -20,10 +20,8 @@ component 'openssl' do |pkg, settings, platform|
 
  if platform.is_windows?
     pkg.environment 'PATH', "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
-  #   pkg.environment 'CYGWIN', settings[:cygwin]
-  #   pkg.environment 'CC', settings[:cc]
-  #   pkg.environment 'CXX', settings[:cxx]
-  #   pkg.environment 'MAKE', platform[:make]
+    pkg.environment 'CYGWIN', settings[:cygwin]
+    pkg.environment 'MAKE', platform[:make]
 
     target = platform.architecture == 'x64' ? 'mingw64' : 'mingw'
   #   cflags = settings[:cflags]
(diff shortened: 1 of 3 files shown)


configs/components/openssl-3.0.rb
```diff
@@ -16,18 +16,18 @@
...
+  if platform.is_windows?
     pkg.environment 'PATH', "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
-  #   pkg.environment 'CYGWIN', settings[:cygwin]
-  #   pkg.environment 'CC', settings[:cc]
-  #   pkg.environment 'CXX', settings[:cxx]
-  #   pkg.environment 'MAKE', platform[:make]
+    pkg.environment 'CYGWIN', settings[:cygwin]
+    pkg.environment 'MAKE', platform[:make]
 
+#    cflags = "-I#{settings[:includedir]}"
```
tvpartytonight, line 28: should these cflag and ldflag assignments be kept around, or do we plan on using them later?

tvpartytonight (approved):
Just a small comment, otherwise looks good to me.

configs/components/openssl-3.0.rb
joshcooper: Ah good catch, I meant to remove that and forgot

joshcooper:
I removed the obsolete comment, but everything else is the same.

PR #673, opened 2023-05-23
(PA-5409) Update Configure flags for openssl3

tvpartytonight:
Small update to the configure flags used with openssl 3.

### DIFF
--- a/configs/components/openssl-3.0.rb
+++ b/configs/components/openssl-3.0.rb
@@ -133,13 +133,26 @@ component 'openssl' do |pkg, settings, platform|
     '--libdir=lib',
     "--openssldir=#{settings[:prefix]}/ssl",
     'shared',
-    'no-asm',
+    'no-gost',
     target,
     sslflags,
     'no-camellia',
     'no-ec2m',
     'no-md2',
-    'no-ssl3'
+    'no-ssl3',
+    'no-ssl3-method',
+    'no-dtls1-method',
+    'no-dtls1_2-method',
+    'no-aria',
+    'no-bf',
+    'no-cast',
+    'no-rc2',
+    'no-rc5',
+    'no-md4',
+    'no-mdc2',
+    'no-rmd160',
+    'no-whirlpool',
+    'no-legacy'
   ]
 
   # Individual projects may provide their own openssl configure flags:


tvpartytonight:
I successfully test building with these flags [here](https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2059/).

configs/components/openssl-3.0.rb
```diff
@@ -133,13 +133,26 @@
...
+    'no-gost',
     target,
     sslflags,
     'no-camellia',
     'no-ec2m',
     'no-md2',
-    'no-ssl3'
+    'no-ssl3',
+    'no-ssl3-method',
+    'no-dtls1-method',
```
joshcooper, line 144:
```suggestion
    'no-dtls',
    'no-dtls1-method',
```

configs/components/openssl-3.0.rb
```diff
@@ -133,13 +133,26 @@
...
+    'no-dtls1_2-method',
+    'no-aria',
+    'no-bf',
+    'no-cast',
```
joshcooper, line 148:
```suggestion
    'no-cast',
    'no-idea',
```

configs/components/openssl-3.0.rb
```diff
@@ -133,13 +133,26 @@
...
+    'no-rc2',
+    'no-rc5',
```
joshcooper, line 150:
```suggestion
    'no-rc5',
    'no-seed',
```

configs/components/openssl-3.0.rb
tvpartytonight: `no-dtls` was already specified [here](https://github.com/tvpartytonight/puppet-runtime/blob/PA-5409/configs/projects/_shared-agent-settings.rb#L229), so I wasn't sure if we wanted to repeat them?

tvpartytonight: I think it is the same for the other suggestions you have; they are already in the `shared-agent-settings.rb`.

joshcooper: ohhh, right I forgot about `openssl_extra_configure_flags`

joshcooper (approved):

PR #677, opened 2023-06-01
Correct URL to libffi source

traylenator:
While this patch is valid its a bit pointless as it is.

Quite a few of these URLs are wrong but it makes little sense to correct them unless they are confirmed as working by the CI.

Internally of course this is not noticed since the local cache'd copy is always used.


traylenator:
CLA signed on behalf of CERN IT with Puppet.

traylenator:
It probably never worked I expect.

PR #678, opened 2023-06-01
(maint) Reenable bf and cast ciphers for pgcrypto

tvpartytonight:
The openssl built for puppet-runtime is used to build postgres for puppetdb; when openssl is built with the `no-bf` and `no-cast` flags, the pgcrypto module for postgres errors out during the build process. This change puts the ciphers back in so postgres can build for puppetdb.

### DIFF
--- a/configs/components/openssl-3.0.rb
+++ b/configs/components/openssl-3.0.rb
@@ -113,8 +113,8 @@ component 'openssl' do |pkg, settings, platform|
     'no-dtls1-method',
     'no-dtls1_2-method',
     'no-aria',
-    'no-bf',
-    'no-cast',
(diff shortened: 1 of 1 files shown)


joshcooper (commented):
Kicked off vanagon build https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-generic-builder-init_generic-builder/2129/ and then will test with `puppet-enterprise-vanagon`

joshcooper (approved):

joshcooper:
I was able to build `bundle exec vanagon build pe-postgresql14-server el-7-x86_64` with this runtime.

PR #683, opened 2023-06-21
(PA-4592) added macos-13-x86_64 platform

amitkarsale:
Added platform for macos13 intel.

### DIFF
--- a/configs/platforms/osx-13-x86_64.rb
+++ b/configs/platforms/osx-13-x86_64.rb
@@ -1,0 +1,8 @@
+platform 'osx-13-x86_64' do |plat|
+  plat.inherit_from_default
+
+  packages = %w[cmake pkg-config yaml-cpp]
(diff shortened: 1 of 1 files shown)


configs/platforms/osx-13-x86_64.rb
```diff
@@ -0,0 +1,8 @@
+platform 'osx-13-x86_64' do |plat|
+  plat.inherit_from_default
```
mhashizume, line 2: Since we're inheriting from default for this platform, we'll need to add it upstream in Vanagon before merging this PR. It looks like this work is ticketed out in PA-5326

amitkarsale: Yess, I have made changes there too. Will raise a PR for the same too.

amitkarsale: @mhashizume  please review this too : https://github.com/puppetlabs/vanagon/pull/807
Thanks!

mhashizume (approved):

PR #680, opened 2023-06-05
(PA-4774) Enable macOS 13 for agent-runtime main

shubhamshinde360:

PR #684, opened 2023-06-26
(PA-5594) Enable macOS 13 (ARM) puppet-runtime builds for agent-runtime-7.x

skyamgarp:
Hello,
The result of component diff :

`
Platform name: osx-13-arm64
    Component 'openssl-1.1.1'
        Field: configure[0]
        --------------------
        - [" ./Configure --prefix=/opt/puppetlabs/puppet --libdir=lib --openssldir=/opt/puppetlabs/puppet/ssl shared no-asm darwin64-x86_64-cc  no-camellia no-ec2m no-md2 no-ssl3 no-dtls no-dtls1 no-idea no-seed no-weak-ssl-ciphers -DOPENSSL_NO_HEARTBEATS -I/opt/puppetlabs/puppet/include "]
        + [" ./Configure --prefix=/opt/puppetlabs/puppet --libdir=lib --openssldir=/opt/puppetlabs/puppet/ssl shared no-asm darwin64-arm64-cc  no-camellia no-ec2m no-md2 no-ssl3 no-dtls no-dtls1 no-idea no-seed no-weak-ssl-ciphers -DOPENSSL_NO_HEARTBEATS -I/opt/puppetlabs/puppet/include "]

        Field: environment
        --------------------
        - PATH=/opt/pl-build-tools/bin:$(PATH):/usr/local/bin
        + PATH=/opt/homebrew/bin:$(PATH):/usr/local/bin

    Component 'ruby-2.7.8'
        Field: configure[0]
        --------------------
        - ["bash configure         --enable-shared         --enable-bundled-libyaml         --disable-install-doc         --disable-install-rdoc                   --prefix=/opt/puppetlabs/puppet --with-opt-dir=/opt/puppetlabs/puppet  --enable-dtrace "]
        + ["bash configure         --enable-shared         --enable-bundled-libyaml         --disable-install-doc         --disable-install-rdoc                   --prefix=/opt/puppetlabs/puppet --with-opt-dir=/opt/puppetlabs/puppet  --with-openssl-dir=/opt/puppetlabs/puppet  --enable-dtrace "]

        Field: environment
        --------------------
        - optflags=-I/opt/puppetlabs/puppet/include CC=clang
        + optflags=-I/opt/puppetlabs/puppet/include CC=clang PATH=/opt/homebrew/bin:$(PATH):/usr/local/bin

        Field: install[12]
        --------------------
        + ["/opt/puppetlabs/puppet/bin/ruby ../rbconfig-update.rb \"{\\\"CC\\\"=>\\\"clang\\\"}\" $$(/opt/puppetlabs/puppet/bin/ruby -e \"puts RbConfig::CONFIG[\\\"topdir\\\"]\")", "cp original_rbconfig.rb /opt/puppetlabs/puppet/share/doc/rbconfig-2.7.8-orig.rb", "cp new_rbconfig.rb $$(/opt/puppetlabs/puppet/bin/ruby -e \"puts RbConfig::CONFIG[\\\"topdir\\\"]\")/rbconfig.rb"]

`

And [ vanagon generic pipeline](https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/1096/)

Also, [Artifacts](http://builds.delivery.puppetlabs.net/puppet-runtime/fb126f661ceba90ecacb689d75e5a1482182fad8/repos/?C=M&O=D)

### DIFF
--- a/configs/components/ruby-2.7.8.rb
+++ b/configs/components/ruby-2.7.8.rb
@@ -111,6 +111,9 @@ component 'ruby-2.7.8' do |pkg, settings, platform|
     # implementation instead of 'arm64', so specify 'amd64' explicitly
     # https://github.com/ruby/ruby/blob/c9c2245c0a25176072e02db9254f0e0c84c805cd/configure.ac#L2329-L2330
     special_flags += " --with-baseruby=#{host_ruby} --with-coroutine=arm64 "
+  elsif platform.is_macos? && platform.architecture == 'arm64' && platform.os_version.to_i >= 13
+    pkg.environment 'PATH', '/opt/homebrew/bin:$(PATH):/usr/local/bin'
+    special_flags += " --with-openssl-dir=#{settings[:prefix]} "
   elsif platform.is_solaris? && platform.architecture == "sparc"
     special_flags += " --with-baseruby=#{host_ruby} --enable-close-fds-by-recvmsg-with-peek "
   elsif platform.name =~ /el-6/
@@ -194,6 +197,7 @@ component 'ruby-2.7.8' do |pkg, settings, platform|
     'x86_64-w64-mingw32' => 'x64-mingw32',
     'i686-w64-mingw32' => 'i386-mingw32'
   }
+
   if target_doubles.key?(settings[:platform_triple])
     rbconfig_topdir = File.join(ruby_dir, 'lib', 'ruby', '2.7.0', target_doubles[settings[:platform_triple]])
   else
@@ -225,6 +229,8 @@ component 'ruby-2.7.8' do |pkg, settings, platform|
       # the ancient gcc version on sles-12-ppc64le does not understand -fstack-protector-strong, so remove the `strong` part
       rbconfig_changes["LDFLAGS"] = "-L. -Wl,-rpath=/opt/puppetlabs/puppet/lib -fstack-protector -rdynamic -Wl,-export-dynamic -L/opt/puppetlabs/puppet/lib"
     end
+  elsif platform.is_macos? && platform.architecture == 'arm64' && platform.os_version.to_i >= 13
+    rbconfig_changes["CC"] = 'clang'
   elsif platform.is_windows?
     rbconfig_changes["CC"] = "x86_64-w64-mingw32-gcc"
   end
--- a/configs/components/yaml-cpp.rb
+++ b/configs/components/yaml-cpp.rb
@@ -28,7 +28,7 @@ component "yaml-cpp" do |pkg, settings, platform|
       pkg.environment 'CXX', 'clang++ -target arm64-apple-macos12' if platform.name =~ /osx-12/
     elsif platform.architecture == 'arm64' && platform.os_version.to_i >= 13
       pkg.environment 'CXX', 'clang++'
-      cmake = "/opt/homebrew/bin/cmake" if platform.architecture == "arm64"
+      cmake = "/opt/homebrew/bin/cmake"
     end
 
   elsif platform.is_windows?
--- a/configs/components/openssl-1.1.1.rb
+++ b/configs/components/openssl-1.1.1.rb
@@ -54,14 +54,19 @@ component 'openssl' do |pkg, settings, platform|
     ldflags = "-R/opt/pl-build-tools/#{settings[:platform_triple]}/lib -Wl,-rpath=#{settings[:libdir]} -L/opt/pl-build-tools/#{settings[:platform_triple]}/lib"
     target = platform.architecture =~ /86/ ? 'solaris-x86-gcc' : 'solaris-sparcv9-gcc'
   elsif platform.is_macos?
-    pkg.environment 'PATH', '/opt/pl-build-tools/bin:$(PATH):/usr/local/bin'
+    if platform.os_version.to_i >= 13 && platform.architecture == 'arm64'
+      pkg.environment 'PATH', '/opt/homebrew/bin:$(PATH):/usr/local/bin'
+    else
+      pkg.environment 'PATH', '/opt/pl-build-tools/bin:$(PATH):/usr/local/bin'
+    end
 
     cflags = settings[:cflags]
-    target = if platform.is_cross_compiled?
-               'darwin64-arm64-cc'
-             else
-               'darwin64-x86_64-cc'
-             end
+
+    target =  if platform.architecture == 'arm64'
+                'darwin64-arm64-cc'
+              else
+                'darwin64-x86_64-cc'
+              end
   elsif platform.is_linux?
     pkg.environment 'PATH', '/opt/pl-build-tools/bin:$(PATH):/usr/local/bin'
 
--- a/configs/platforms/osx-13-arm64.rb
+++ b/configs/platforms/osx-13-arm64.rb
@@ -1,9 +1,8 @@
 platform 'osx-13-arm64' do |plat|
-    plat.inherit_from_default
-  
-    packages = %w[automake cmake pkg-config yaml-cpp]
-    plat.provision_with "su test -c '/opt/homebrew/bin/brew install #{packages.join(' ')}'"
-  
-    plat.output_dir File.join('apple', '13', 'PC1', 'arm64')
-  end
-  
+  plat.inherit_from_default
+
+  packages = %w[automake cmake pkg-config yaml-cpp]
+  plat.provision_with "su test -c '/opt/homebrew/bin/brew install #{packages.join(' ')}'"
+
+  plat.output_dir File.join('apple', '13', 'PC1', 'arm64')
+end


configs/components/_base-ruby-augeas.rb
```diff
@@ -20,6 +20,11 @@
 pkg.build_requires "augeas"
 
 pkg.environment "PATH", "$(PATH):/opt/pl-build-tools/bin:/usr/local/bin:/opt/csw/bin:/usr/ccs/bin:/usr/sfw/bin"
+
+if platform.architecture == 'arm64' && platform.is_macos? && platform.os_version.to_i >= 13
+  pkg.environment "PATH", "/opt/homebrew/bin:$(PATH):/opt/pl-build-tools/bin:/usr/local/bin:/opt/csw/bin:/usr/ccs/bin:/usr/sfw/bin"
```
tvpartytonight, line 25: I think that here you don't need to repeat `/opt/pl-build-tools/bin:/usr/local/bin:/opt/csw/bin:/usr/ccs/bin:/usr/sfw/bin`. You can just do `"/opt/homebrew/bin:$(PATH)"`, similar to how [this line](https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/_base-ruby-augeas.rb#L29) is written.

configs/components/_base-ruby.rb
```diff
@@ -50,7 +50,7 @@
   pkg.environment 'optflags', settings[:cflags] + ' -O3'
 elsif platform.is_macos?
   pkg.environment 'optflags', settings[:cflags]
-  if platform.is_cross_compiled?
+  if platform.architecture == 'arm64'
     # Pin to an older version of ruby@2.5. This can be removed once we're no longer cross-compiling
```
tvpartytonight, line 54: I think we should be able to remove this conditional in pinning to ruby 2.5?

configs/components/ruby-2.7.8.rb
```diff
@@ -203,19 +207,21 @@
   rbconfig_changes = {}
   if platform.is_aix?
     rbconfig_changes["CC"] = "gcc"
-  elsif platform.is_cross_compiled? || platform.is_solaris?
+  elsif platform.is_macos? && platform.architecture == 'arm64'
     if platform.name =~ /osx-11/
```
tvpartytonight, line 211: I think that these `osx-11` and `osx-12` conditions need to be moved down inside the `elsif platform.is_cross_compiled? || platform.is_solaris?` block.

configs/components/yaml-cpp.rb
```diff
@@ -23,9 +23,14 @@
   elsif platform.is_macos?
     cmake_toolchain_file = ""
     cmake = "/usr/local/bin/cmake"
-    if platform.is_cross_compiled?
+    if platform.architecture == 'arm64'
```
tvpartytonight, line 26: I think you will still need the `platform.is_cross_compiled?` condition here because the conditions for `osx-11` and `osx-12` won't be executed for x86_64.

configs/projects/_shared-agent-settings.rb
```diff
@@ -200,7 +200,7 @@
   # break gcc or clang if they try to use the RPATH values we forced.
   proj.setting(:cppflags, "-I#{proj.includedir}")
   proj.setting(:ldflags, "-L#{proj.libdir}")
-  if platform.is_cross_compiled?
+  if platform.architecture == 'arm64'
```
tvpartytonight, line 203: I had a similar comment about this [here](https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1242927561).

configs/components/_base-ruby.rb
skyamgarp: But we still want to cross-compile for osx-11 and osx-12

configs/components/ruby-2.7.8.rb
skyamgarp: I think this is fine as the target will get set for 11 & 12 as well once they pass this condition. I will confirm once by running build for them as well on x86_64.

configs/components/yaml-cpp.rb
skyamgarp: Hello,
I verified on this one that architecture will be 'arm64' though we are building on x86_64 as we mention the build tatget to be `osx-12-arm64` for particular project. Please let me know in case of any concern I missed here.

configs/projects/_shared-agent-settings.rb
skyamgarp: As Josh said this is under `platform.is_macos??` condition so we should be good.
https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1248223570

joshcooper:
Since macOS 13 ARM and Intel share many of the same components, e.g. augeas, ruby, etc, we're having to make the same set of changes in this PR and https://github.com/puppetlabs/puppet-runtime/pull/680. For example, see my comment about cross-compiling in https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1253820661. 

I talked with @shubhamshinde360 yesterday and I think we're close to landing his PR. Let's focus on getting that merged and then rebase this PR against upstream/master. I'll also leave some specific comments inline.

configs/components/_base-ruby-augeas.rb
joshcooper: We can exclude this change after rebasing, see https://github.com/puppetlabs/puppet-runtime/pull/680/files#diff-e9d1d8770ae2de7f9bd9522753b5be0e02ce8b3020b57266e7f5d89ff4766f19

configs/components/_base-ruby.rb
joshcooper: We can exclude this change after rebasing, see https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1253820661. 

Also @tvpartytonight is right that Ruby 2.5 is no longer needed (it was used in Puppet 6 which is EOL). If there isn't a PA ticket filed already about removing 6.x code from the runtime, can you file one and we can delete that code later?

configs/components/ruby-2.7.8.rb
```diff
@@ -203,19 +207,21 @@
...
       rbconfig_changes["CC"] = 'clang -target arm64-apple-macos12'
-    else
-      rbconfig_changes["CC"] = "gcc"
-      if platform.is_solaris?
-        # this should probably be done for aix and cross compiled targets too
-        rbconfig_changes["CPP"] = "gcc -E"
-      end
-      rbconfig_changes["warnflags"] = "-Wall -Wextra -Wno-unused-parameter -Wno-parentheses -Wno-long-long -Wno-missing-field-initializers -Wno-tautological-compare -Wno-parentheses-equality -Wno-constant-logical-operand -Wno-self-assign -Wunused-variable -Wimplicit-int -Wpointer-arith -Wwrite-strings -Wdeclaration-after-statement -Wimplicit-function-declaration -Wdeprecated-declarations -Wno-packed-bitfield-compat -Wsuggest-attribute=noreturn -Wsuggest-attribute=format -Wno-maybe-uninitialized"
+    elsif platform.name =~ /osx-13/
+      rbconfig_changes["CC"] = 'clang -target arm64-apple-macos13'
```
joshcooper, line 216: Similar comment here as in https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1253820661. Since we're native compiling for macOS 13 ARM, we shouldn't need to specify the `target` and instead just specify the compiler, similar to the AIX case earlier on:

```suggestion
      rbconfig_changes["CC"] = 'clang'
```

configs/components/ruby-2.7.8.rb
joshcooper: It might be less confusing to do something like https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1253820661

```ruby
elsif platform.is_macos?
  if platform.is_cross_compiled?
    rbconfig_changes["CC"] = 'clang -target arm64-apple-macos11' if platform.name =~ /osx-11/
    rbconfig_changes["CC"] = 'clang -target arm64-apple-macos12' if platform.name =~ /osx-12/
  else
    rbconfig_changes["CC"] = 'clang'
  end
elsif platform.is_solaris?
  ...
```

configs/components/yaml-cpp.rb
```diff
@@ -23,9 +23,14 @@
...
       pkg.environment 'CXX', 'clang++ -target arm64-apple-macos11' if platform.name =~ /osx-11/
       pkg.environment 'CXX', 'clang++ -target arm64-apple-macos12' if platform.name =~ /osx-12/
+      pkg.environment 'CXX', 'clang++ -target arm64-apple-macos13' if platform.name =~ /osx-13/
+
+      if platform.os_version.to_i >= 13
+        cmake = "/opt/homebrew/bin/cmake"
+      end
```
joshcooper, line 33: I left a comment about this in https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1253955838

configs/projects/_shared-agent-settings.rb
joshcooper: :+1: yes that makes sense, left a similar comment in https://github.com/puppetlabs/puppet-runtime/pull/680/files#r1248223570

joshcooper:
@skyamgarp I was able to reproduce the build failure. The issue is a combination of brew bumping packages to depend on openssl 3 and the fact that we're doing native compiles on macOS 13 ARM.

First, the build fails because it can't find `OpenSSL >= 1.0.1, < 3.0.0`, note the "strictly less than 3"

```
openssl:
	Could not be configured. It will not be installed.
	/private/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.n60vh4wY/ruby-2.7.8/ext/openssl/extconf.rb:111: OpenSSL >= 1.0.1, < 3.0.0 or LibreSSL >= 2.5.0 is required
	Check ext/openssl/mkmf.log for more details.
```

Next, we can see brew installed OpenSSL 3:

```
pix-arm64-macos13-1:tmp.n60vh4wY root# cd /Users/test/
pix-arm64-macos13-1:test root# su test /opt/homebrew/bin/brew --prefix openssl
/opt/homebrew/opt/openssl@3
pix-arm64-macos13-1:test root# /opt/homebrew/opt/openssl@3/bin/openssl version
OpenSSL 3.1.1 30 May 2023 (Library: OpenSSL 3.1.1 30 May 2023)
```

This is due to perl's dependencies:

```
09:32:38 su test -c '/usr/local/bin/brew install perl autoconf automake libtool'
...
09:32:42 ==> Fetching dependencies for perl: ca-certificates, openssl@3, berkeley-db and gdbm
```

This can be fixed by passing `--with-openssl-dir=/opt/puppetlabs/puppet` in the ruby configure script, so that ruby uses the openssl we just built earlier on in the build (as opposed to the one brew installed):

```sh
cd ruby-2.7.8 && \
        bash configure --enable-shared --enable-bundled-libyaml --disable-install-doc --disable-install-rdoc --prefix=/opt/puppetlabs/puppet --with-opt-dir=/opt/puppetlabs/puppet --with-baseruby=/opt/puppetlabs/puppet/bin/ruby --with-coroutine=arm64 --enable-dtrace --with-openssl-dir=/opt/puppetlabs/puppet
```

This doesn't break the `agent-runtime-main` build on macOS 13 ARM, because we're building openssl 3.2.2, which is "close enough" to the version brew installed.

And I think `agent-runtime-7.x` on macOS 12 Intel doesn't have this issue, because we're getting lucky. We don't have a `build_requires` dependency on `perl`  so we just use openssl 1.1.x.

Long story short, I think we should be adding `--with-openssl-dir=/opt/puppetlabs/puppet` on all ruby native compilation builds. I'd suggest adding to just the ruby-2.7.8 component on macOS 13 ARM and then later add it to other native macOS builds for ruby-2.7.8 and ruby-3.2.2.

Also for posterity, homebrew seems to require ruby >= 2.6.10, because it only installs it on macOS 11 and 12, but not 13. And those have different preinstalled ruby versions:

```
$ for mac in alma-decency.delivery.puppetlabs.net skimpy-framer.delivery.puppetlabs.net reflective-cure.vmpooler-prod.puppet.net; do echo && echo $mac && ssh $mac 'sw_vers && ruby --version'; done

alma-decency.delivery.puppetlabs.net
ProductName:	macOS
ProductVersion:	11.2.1
BuildVersion:	20D75
ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin20]

skimpy-framer.delivery.puppetlabs.net
ProductName:	macOS
ProductVersion:	12.2.1
BuildVersion:	21D62
ruby 2.6.8p205 (2021-07-07 revision 67951) [universal.x86_64-darwin21]

reflective-cure.vmpooler-prod.puppet.net
ProductName:		macOS
ProductVersion:		13.3.1
ProductVersionExtra:	(a)
BuildVersion:		22E772610a
ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.x86_64-darwin22]
```

configs/components/ruby-2.7.8.rb
```diff
@@ -106,7 +106,10 @@
     special_flags += " --build=#{settings[:platform_triple]} "
   elsif platform.is_cross_compiled? && platform.is_linux?
     special_flags += " --with-baseruby=#{host_ruby} "
-  elsif platform.is_cross_compiled? && platform.is_macos?
+  elsif platform.is_macos? && platform.architecture == 'arm64'
+    if platform.os_version.to_i >= 13
+      pkg.environment 'PATH', '/opt/homebrew/bin:$(PATH):/usr/local/bin'
```
joshcooper, line 111: Suggest adding this for macOS 13+ ARM only for now:

```ruby      
      special_flags += " --with-openssl-dir=#{settings[:prefix]} "
```

And then file a ticket to investigate adding to all native builds later:

```ruby
special_flags = " --prefix=#{ruby_dir} --with-opt-dir=#{settings[:prefix]} "
special_flags += "--with-openssl-dir=#{settings[:prefix]} " unless platform.is_cross_compiled?
```

joshcooper (approved):

PR #695, opened 2023-07-06
(PA-5617) Build yaml-cpp on AIX 7.2

joshcooper:
Previous attempts to build yaml-cpp failed when trying to run
/opt/freeware/bin/cmake, because it linked to libcurl.a and generated:

    0509-150 Dependent module /opt/freeware/lib/libcurl.a(libcurl.so.4) could not be loaded.
    0509-103 The module has an invalid magic number.

That library is preinstalled on our AIX 7.2 hosts and is version 7.52.1. The
library is also used by yum/python in order to download AIX toolbox packages.

As mentioned in https://github.com/eclipse-openj9/openj9/issues/4245 upgrading
the curl version resolves the issue. And in order to do that we need to bootstrap
a newer openssl 1.1.x version.

Installing newer curl also means we can remove the hack that was added when
building curl.

Note the typo `$PATH` vs `$(PATH)`. The latter is necessary so that PATH is
resolved at runtime when make executes. Sometimes it's also written as `$$PATH`.

Built in https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/BUILD_TARGET=aix-7.2-ppc,SLAVE_LABEL=k8s-worker/2115/

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -41,17 +41,6 @@ component 'curl' do |pkg, settings, platform|
     configure_options << "--disable-dependency-tracking"
   end
 
-  if platform.name == 'aix-7.2-ppc'
-    # yum on aix installs an old version of libcurl.a that /opt/freeware/bin/gcc seems
-    # to use no matter what -L search path I use, so use the same workaround as bbf248fb6
-    pkg.configure do
-      [
-        'mkdir -p /opt/freeware/lib/hide',
-        'mv /opt/freeware/lib/libcurl.a /opt/freeware/lib/hide/libcurl.a'
-      ]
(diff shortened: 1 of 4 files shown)


joshcooper:
```
❯ bx rake vanagon:component_diff -- -P agent-runtime-main -p all --from 46c127e --to HEAD  | grep 'Platform name' 
Platform name: aix-7.2-ppc

❯ bx rake vanagon:component_diff -- -P agent-runtime-7.x -p all --from 46c127e --to HEAD  | grep 'Platform name'
Platform name: aix-7.2-ppc
```

PR #698, opened 2023-07-12
(PA-5640) Fix ruybgem-ffi runtime logic

cthorn42:
Our logic is incorrect, previously we were looking at `settings[:runtime_project]`. In agent-runtime context that will always just return 'agent', not the full project name we were looking for.
This lead to having the PL build tools path added earlier then we want and an older GCC being used that breaks rubygem-ffi from building.


cthorn42:
Link to vanagon generic builder for agent-runtime-main with these changes for solaris-i386: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2133/

joshcooper (approved):
Ah good catch! Also wondering if we should skip `pkg.install_file` when running on Ruby 3. But that's a tomorrow problem.

PR #700, opened 2023-07-17
(PA-5640) Fix up Solaris' ffi gem

cthorn42:
This PR cleans up a bit of the Solaris FFI gem work around Solaris. We no longer copy over the system libffi and rely on the OpenCSW libffi we install. Also when installing them ffi gem we use the OpenCSW libffi.

### DIFF
--- a/configs/components/_base-rubygem.rb
+++ b/configs/components/_base-rubygem.rb
@@ -21,6 +21,9 @@ if platform.is_windows?
   pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(shell cygpath -u #{settings[:ruby_bindir]}):$(shell cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0:$(PATH)"
 end
 
+if name =~ /ffi/ && settings[:ruby_version].to_f >= 3.2 && platform.is_solaris?
+  extra_install_settings = ' -- --enable-system-libffi'
+end
 # When cross-compiling, we can't use the rubygems we just built.
 # Instead we use the host gem installation and override GEM_HOME. Yay?
 pkg.environment "GEM_HOME", settings[:gem_home]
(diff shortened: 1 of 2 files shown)


configs/components/_base-rubygem.rb
```diff
@@ -21,6 +21,9 @@
   pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(shell cygpath -u #{settings[:ruby_bindir]}):$(shell cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0:$(PATH)"
 end
 
+if name =~ /ffi/ && settings[:ruby_version].to_f >= 3.2 && platform.is_solaris?
+  extra_install_settings = ' -- --enable-system-libffi'
+end
```
joshcooper, line 26: The `libffi` thing has confused me for awhile, but I think I'm starting to understand. 

Since `libffi` is a runtime component (as of Ruby 3.2), then we're [building and installing it](https://github.com/puppetlabs/puppet-runtime/blob/d31b127e27f5761658d473f2d13798c079da5648/configs/components/libffi.rb#L48-L57) as part of the runtime build. So when components that depend on `libffi`, like `ruby`, etc, we should (in general) build against the same version built.

For example, on Solaris 11 Intel, we're installing [libffi-dev](https://github.com/puppetlabs/puppet-runtime/blob/d31b127e27f5761658d473f2d13798c079da5648/configs/platforms/solaris-11-i386.rb#L44) from OpenCSW, currently [version 3.2.1](https://www.opencsw.org/packages/libffi_dev/) and we compile against that. But we ship [libffi version 3.4.3](https://github.com/puppetlabs/puppet-runtime/blob/d31b127e27f5761658d473f2d13798c079da5648/configs/components/libffi.rb#L2). Those differences could introduce runtime failures.

Ideally, I think we want to specify the `libffi` dependency, something like:

```
gem install ffi-1.15.5.gem --local -- --with-libffi-dir=/opt/puppetlabs/puppet
```

I think a better option is to tell the gem to use `pkg-config` to resolve dependencies, something like:

```
PKG_CONFIG_PATH=/opt/puppetlabs/puppet/lib/pkgconfig:$PKG_CONFIG_PATH gem install ffi-1.15.5.gem --local
```

Similar to what's being done in https://github.com/ffi/ffi/blob/a480ff6db209713fc73a3dbbb4f1c31a376d6e6a/.github/workflows/ci.yml#L34

I think this might solve the issue @AriaXLi was seeing with the ffi gem on macOS 12 ARM as well.

cthorn42: I'm trying out the `PKG_CONFIG_PATH` suggestion, if that builds and completes I'll close this and put up a new PR with that work.

PR #703, opened 2023-07-19
(maint) Add DES back to OpenSSL 3

justinstoller:
We ship pgcrypto in PE and it uses the agent runtime's openssl. pgcrypto hardcodes DES support into it and will fail to start if the DES algorithm isn't available.

In the future either postgresql should ship its own openssl or we should patch pgcrypto to not hardcode DES into itself.

### DIFF
--- a/configs/components/openssl-3.0.rb
+++ b/configs/components/openssl-3.0.rb
@@ -127,6 +127,9 @@ component 'openssl' do |pkg, settings, platform|
     'no-aria',
     # 'no-bf', pgcrypto is requires this cipher in postgres for puppetdb
     # 'no-cast', pgcrypto is requires this cipher in postgres for puppetdb
+    # 'no-des', pgcrypto is requires this cipher in postgres for puppetdb,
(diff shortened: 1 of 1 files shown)


donoghuc (approved):
Whack a mole

configs/components/openssl-3.0.rb
```diff
@@ -136,7 +136,7 @@
   if settings[:use_legacy_openssl_algos]
     pkg.apply_patch 'resources/patches/openssl/openssl-3-activate-legacy-algos.patch'
   else
-    configure_flags << 'no-legacy' << 'no-md4' << 'no-des'
+    configure_flags << 'no-legacy' << 'no-md4'
```
joshcooper, line 139: Can you add a `no-des` (commented out) like we're doing for `no-bf` so we know why it's not being disabled.

justinstoller: Updated with info about pgcrypto and ntlm

joshcooper (approved):

PR #705, opened 2023-07-20
(PE-36344) Stop using fork of ntlm gem

donoghuc:
We are now shipping openssl runtimes with the required crypto algorithms. This commit stops using the ntlm fork that provided a ruby implementation for the crypto algs.


configs/projects/_shared-pe-bolt-server_with_ruby.rb
```diff
@@ -176,6 +176,7 @@
 proj.component('rubygem-net-ssh')
 proj.component('rubygem-net-ssh-krb')
 proj.component('rubygem-nori')
+proj.component('rubygem-rubyntlm')
```
justinstoller, line 179: Oh, when I reverted the usage of the ntlm fork in pe-installer-runtime-main (d9934091496a68bb4bdf3f1b07ce67cc65d852ff) I didn't see a dependency on 'rubygem-rubyntlm' added. Does something similar need to be done to pe-instlaler?

configs/projects/pe-bolt-server-runtime-main.rb
```diff
@@ -11,9 +11,6 @@
   proj.setting(:openssl_version, '3.0')
 
   instance_eval File.read(File.join(File.dirname(__FILE__), '_shared-pe-bolt-server_with_ruby.rb'))
-
-  # TODO: Work around PE-36078 by using forked non-optimal solution
-  proj.component('rubygem-rubyntlm-fork')
```
justinstoller, line 16: whoops! I should have removed this when I added the `proj.setting(:use_legacy_openssl_algos, true)` in the bolt server, sorry.

justinstoller (approved):
Looks great, just have a quick question if this has further implications for pe-installer-runtime-main

configs/projects/_shared-pe-bolt-server_with_ruby.rb
donoghuc: Not now, when i was looking at this i realized for installer we actually use rubygems to do dependency resolution. This is not ideal because https://tickets.puppetlabs.com/browse/PE-36494 We will explicitly package that when we do that ticket but for now rubygems will find the latest rubyntmlm and we will be good to go.

configs/projects/pe-bolt-server-runtime-main.rb
donoghuc: ah, yeah i'll add it

justinstoller: oh, lol, i hadn't noticed that Sean didn't include my update for `:use_legacy_openssl_algos`. Thanks, Cas!

PR #708, opened 2023-08-01
(PA-5666) Fix pl-ruby-patch's filename

cthorn42:
Starting with Ruby 3.2.2 we want to change the file we are sed'ing.

### DIFF
--- a/configs/components/pl-ruby-patch.rb
+++ b/configs/components/pl-ruby-patch.rb
@@ -61,9 +61,16 @@ component "pl-ruby-patch" do |pkg, settings, platform|
       sed_command = %(s|Gem.ruby.shellsplit|& << '-r/opt/puppetlabs/puppet/share/doc/rbconfig-#{settings[:ruby_version]}-orig.rb'|)
     end
 
+    # rubygems switched which file has the command we need to patch starting with ruby 3.2.2
+    if Gem::Version.new(settings[:ruby_version]) >= Gem::Version.new('3.2.2')
+      filename = 'builder.rb'
+    else
+      filename = 'ext_conf_builder.rb'
+    end
+
     pkg.build do
       [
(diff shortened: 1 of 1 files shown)


cthorn42:
Example adhoc job built with this runtime passing with osx-12-arm: https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1158/

configs/components/pl-ruby-patch.rb
```diff
@@ -61,9 +61,16 @@
       sed_command = %(s|Gem.ruby.shellsplit|& << '-r/opt/puppetlabs/puppet/share/doc/rbconfig-#{settings[:ruby_version]}-orig.rb'|)
     end
 
+    # rubygems switched which file has the command we need to patch starting with ruby 3.2.2
+    if Gem::Version.new(settings[:ruby_version]) >= Gem::Version.new('3.2.2')
```
cthorn42, line 65: I'm not sure of a better way to check for semantic versioning, but this is what a Google search suggested and it works.

joshcooper:
This this the commit to ruby's `ruby_3_2` branch that bumped rubygems and changed the ext_conf_builder.rb path: https://github.com/ruby/ruby/commit/c3c461c4ff1b998ac49ed1a3ad411d2f0a9123b8#diff-302705c03d7ce33c91c6f9496f9fafc222984d40b34cb2896118ec248f1575f2

configs/components/pl-ruby-patch.rb
joshcooper: :+1: we use that in puppet too

joshcooper (approved):

PR #713, opened 2023-08-09
(PE-36344) Enable md4 for winrm transport in bolt server for non fips

donoghuc:
This commit enables legacy openssl algorithms to support the winrm transport in bolt-server. Winrm does not work on FIPS currently so we continue to leave that unsupported.


justinstoller:
This also should remove the value you put into pe-installer-runtime-main

donoghuc:
@justinstoller Sorry i dont understand your comment. Are you asking if we need to disable legacy algos on fips for instaler runtime as well?

donoghuc:
Ahhhh, ok I see. Thanks Justin!

jpartlow (approved):
:+1: Enables, except for fips.

PR #723, opened 2023-08-31
(PA-5647) Remove code Changes of Puppet 6.x from puppet-runtime

shubhamshinde360:
This commit removes
1) The project file for agent runtime 6.x - configs/projects/agent-runtime-6.x.rb
2) Projects which satisfy *-2019.8.x.rb
3) Files which refer to ruby version 2.5.9
4) Removes patches for ruby 2.5.9 which are only used by ruby-2.5.9


mhashizume (approved):
I build agent-runtime-7.x in the Vanagon generic builder with some select platforms and did not have any issues. I'm good to merge now and just roll back if this introduces any issues in CI.

PR #737, opened 2023-10-10
(PA-5632) Update SELinux on Red Hat 9 to 3.3

mhashizume:
Prior to this commit, there was a discrepancy between the versions of SELinux we built our Red Hat Enterprise Linux (RHEL) 9 runtimes on and the minimum version available on the OS (SELinux 3.3).

This commit builds RHEL 9 runtimes with version 3.3 of SELinux to resolve that discrepancy.

### DIFF
--- a/configs/components/_base-ruby-selinux.rb
+++ b/configs/components/_base-ruby-selinux.rb
@@ -10,23 +10,34 @@ ruby_version ||= settings[:ruby_version]
 host_ruby ||= settings[:host_ruby]
 ruby_bindir ||= settings[:ruby_bindir]
 
+# We download tarballs because system development packages (e.g.
+# libselinux-devel) don't necessarily include Swig interface files (*.i files)
 if platform.name =~ /el-(6|7)|ubuntu-(16|18.04-amd64)/
   pkg.version "2.0.94"
   pkg.md5sum "544f75aab11c2af352facc51af12029f"
   pkg.url "https://raw.githubusercontent.com/wiki/SELinuxProject/selinux/files/releases/20100525/devel/libselinux-#{pkg.get_version}.tar.gz"
-  pkg.mirror "#{settings[:buildsources_url]}/libselinux-#{pkg.get_version}.tar.gz"
+elsif platform.name.start_with?('el-9')
+  # SELinux 3.3 is the minimum version available in RHEL 9 repos
+  pkg.version '3.3'
+  pkg.sha256sum 'acfdee27633d2496508c28727c3d41d3748076f66d42fccde2e6b9f3463a7057'
+  pkg.url "https://github.com/SELinuxProject/selinux/releases/download/#{pkg.get_version}/libselinux-#{pkg.get_version}.tar.gz"
 else
   pkg.version "2.9"
   pkg.md5sum "bb449431b6ed55a0a0496dbc366d6e31"
   pkg.apply_patch "resources/patches/ruby-selinux/selinux-29-function.patch"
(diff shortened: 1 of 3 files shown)


mhashizume:
These changes were built and passed an ad hoc pipeline for both x86_64 and aarch64:

https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=redhat9-64a/1194/
https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=redhat9-AARCH64a/1194/

This is definitely pretty hacky so feedback and improvements welcome

configs/components/_base-ruby-selinux.rb
```diff
@@ -57,6 +68,11 @@
     steps << "#{platform.patch} --strip=0 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../selinuxswig_ruby_wrap.patch"
   end
 
+  # libselinux 3.3 is the minimum version we want to build on RHEL 9, but the
```
tvpartytonight, line 71: Have we tried using versions 3.4 or 3.5? I'm fine merging this as is, but, if we could just use a later version of selinux, maybe we can avoid this hack?

mhashizume: I haven't tried it, my assumption was that there would be compatibility issues if there were a discrepancy between the bindings we built and the version running on the system. Do you think it would be worth a shot?

tvpartytonight: Yeah, I'm not sure if there would be a discrepancy, seems like it might possibly work? As long as the major versions are the same...

mhashizume: I tried using the SELinux 3.4 tarball and installing the libselinux-devel-3.4 package. It seems like it creates the same compatibility issue as we saw in the original ticket. On an adhoc agent build on x86_64:
```
Begin tests/selinux.rb

PA-3067: Manage selinux

  * test require 'selinux'

  ergotropic-trap.delivery.puppetlabs.net (ergotropic-trap.delivery.puppetlabs.net) 23:22:11$ env PATH="/opt/puppetlabs/puppet/bin:${PATH}" ruby -e 'require "selinux"'
 /opt/puppetlabs/puppet/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:86:in `require': /lib64/libselinux.so.1: version `LIBSELINUX_3.4' not found (required by /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/2.7.0/x86_64-linux/selinux.so) - /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/2.7.0/x86_64-linux/selinux.so (LoadError)
    	from /opt/puppetlabs/puppet/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:86:in `require'
    	from -e:1:in `<main>'
```
I assume this happens because there's a mismatch between what we build against and what's present on the system.

I think the only way we can avoid this issue is to target the lowest common denominator, SELinux 3.3.

tvpartytonight (approved):

PR #738, opened 2023-10-11
(PA-5852) Apply CVE-2023-38545 patch for curl vulnerablity

amitkarsale:
apply patch for curl CVE-2023-38545

ref: https://github.com/curl/curl/discussions/12026
https://curl.se/docs/CVE-2023-38545.html

### DIFF
--- a/resources/patches/curl/CVE-2023-38545.patch
+++ b/resources/patches/curl/CVE-2023-38545.patch
@@ -1,0 +1,27 @@
+diff --git a/lib/socks.c b/lib/socks.c
+index 95c2b004c..8cf694d1d 100644
+--- a/lib/socks.c
++++ b/lib/socks.c
+@@ -588,9 +588,9 @@ static CURLproxycode do_SOCKS5(struct Curl_cfilter *cf,
+ 
+     /* RFC1928 chapter 5 specifies max 255 chars for domain name in packet */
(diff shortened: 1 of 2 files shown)


resources/patches/curl/CVE-2023-38545.patch
```diff
@@ -0,0 +1,110 @@
...
+       else {
+         socksreq[len++] = 3;
+-        socksreq[len++] = (char) hostname_len; /* one byte address length */
++        socksreq[len++] = (unsigned char) hostname_len; /* one byte length */
+         memcpy(&socksreq[len], sx->hostname, hostname_len); /* w/o NULL */
+         len += hostname_len;
+       }
+diff --git a/tests/data/Makefile.inc b/tests/data/Makefile.inc
+index 97fdabcfa..f713c03b3 100644
+--- a/tests/data/Makefile.inc
```
mhashizume, line 29: Do we need to bring in test files as well?

amitkarsale: As per @joshcooper's comment on chat I have updated the PR to omit the test file.

cthorn42 (approved):
Ran the two agent runtime builds through with the recent changes that dropped testing and they built fine. We're good to go.

PR #739, opened 2023-10-13
(CAT-270) Set special flag to run against Mac OS on ruby 3.2.2

david22swan:
Jenkins builds failing for Mac OSX with current version: https://jenkins-platform.delivery.puppetlabs.net/job/platform_pdk-runtime_runtime-vanagon-packaging_main/BUILD_TARGET=osx-12-x86_64,SLAVE_LABEL=k8s-worker/314/console

Set special flag `--with-openssl-dir=` to run against Mac OS on Ruby 3.2.2.
This flag works to force the use of the openssl version included within the runtime itself rather than the one that is installed by homebrew as default.

### DIFF
--- a/configs/components/ruby-3.2.2.rb
+++ b/configs/components/ruby-3.2.2.rb
@@ -129,6 +129,8 @@ component 'ruby-3.2.2' do |pkg, settings, platform|
     else
       special_flags += " --build i686-w64-mingw32 "
     end
+  elsif platform.is_macos?
+    special_flags += " --with-openssl-dir=#{settings[:prefix]} "
   end
 
   without_dtrace = [


david22swan:
Screen Shot 2023-10-19 at 9 16 20 AM

LukasAud (approved):
LGTM

gavindidrichsen (approved):
Well Done!

joshcooper (approved):
Not a blocking comment, just kind of surprised we're not already doing this for all platforms. For example, we do the same when building `curl` for all platforms: https://github.com/puppetlabs/puppet-runtime/blob/192c686301b43eb2427af6f9e2084634a2435514/configs/components/curl.rb#L36

gavindidrichsen:
Thanks @joshcooper!

PR #755, opened 2023-11-02
(PA-5632) Check for SELinux symlink

mhashizume:
Prior to this commit, when building runtimes the SELinux component created a symlink for a shared library on Red Hat 9 in order to successfully build SELinux 3.3.

This worked for the agent runtimes but because the PDK runtime builds multiple Rubies, builds would fail because it would attempt to create a symlink when one already existed.

This commit checks for a symlink for the SELinux shared library before creating one.


mhashizume:
- [ ] PDK build
- [ ] Agent build

joshcooper (approved):

PR #756, opened 2023-11-06
(CAT-270) Set unique dirs for OpenSSL versions on mac

david22swan:


david22swan:
Curl change gives error: `configure: error: /opt/puppetlabs/pdk/openssl-1.1.1/ssl is a bad --with-openssl prefix!`

PR #758, opened 2023-11-09
Build pdk runtime components against openssl 3

joshcooper:
The order that components are listed in pdk-components matters, so build openssl 3 first followed by other components that rely on it (curl, git, etc). Then build openssl 1.1.1 before building the extra ruby 2.7.8.

Note the openssl 1.1.1 headers still overwrite the 3.0.x, which is bad and needs to be resolved somehow as it may confuse any rubygem with native extension that we install in ruby 2.7.

```
ashen-control:~ root# otool -L /opt/puppetlabs/pdk/lib/libcurl.4.dylib
/opt/puppetlabs/pdk/lib/libcurl.4.dylib:
	/opt/puppetlabs/pdk/lib/libcurl.4.dylib (compatibility version 13.0.0, current version 13.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1856.105.0)
	/System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1163.60.3)
	/opt/puppetlabs/pdk/lib/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/opt/puppetlabs/pdk/lib/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
ashen-control:~ root# export PDK_DISABLE_ANALYTICS="true"
ashen-control:~ root# export PATH="/opt/puppetlabs/pdk/bin:$PATH"
ashen-control:~ root# /opt/puppetlabs/pdk/private/git/bin/git clone https://github.com/puppetlabs/puppetlabs-motd /var/root/puppetlabs-motd
Cloning into '/var/root/puppetlabs-motd'...
remote: Enumerating objects: 2366, done.
remote: Counting objects: 100% (579/579), done.
remote: Compressing objects: 100% (259/259), done.
remote: Total 2366 (delta 297), reused 474 (delta 279), pack-reused 1787
Receiving objects: 100% (2366/2366), 549.52 KiB | 16.16 MiB/s, done.
Resolving deltas: 100% (1172/1172), done.
```

```
ashen-control:~ root# /opt/puppetlabs/pdk/private/ruby/2.7.8/bin/irb
[...]

PR #717, opened 2023-08-15
(maint) Add some more developer docs

joshcooper:
* Windows
* Cross-compiling

PR #759, opened 2023-11-13
(maint) Add design decision for recent pdk-runtime PR #758

gavindidrichsen:
This PR is an attempt to answer why and how this [pdk-runtime PR](https://github.com/puppetlabs/puppet-runtime/pull/758) attempted to address a critical issue stemming from a homebrew update causing a version mismatch in OpenSSL libraries and headers within pdk-runtime.

As there seems to be more potential changes required to compensate for the impact of the homebrew update I'm moving this to Draft until testing is complete.

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+doc/architecture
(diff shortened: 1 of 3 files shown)


joshcooper:
I'm generally in favor of documenting why things are the way we are for the next poor soul that has to work on the runtime. For example, I have this PR up https://github.com/puppetlabs/puppet-runtime/pull/717

I think my only question is we already have a `notes` directory, so maybe put the adr stuff under `notes/architecture`? Or rename `notes` to `doc`? Having both directories seems not great.

PR #761, opened 2023-11-14
(CAT-1505) Update Windows values for when using OpenSSL 3

david22swan:
Updating the code so that the `ssl_lib` and `ssl_postfix` are pointing to the correct location when OpenSSL 3.0 is being used.


david22swan:
@joshcooper Could you take a look if you have a moment?

joshcooper (approved):

PR #765, opened 2023-11-18
(maint) only force libxml2/libxslt internal mirror if necessary

h0tw1r3:
unmolested source mirror should be added to artifactory

### DIFF
--- a/configs/components/libxml2.rb
+++ b/configs/components/libxml2.rb
@@ -1,12 +1,7 @@
 component "libxml2" do |pkg, settings, platform|
   pkg.version '2.10.3'
-  pkg.sha256sum '26d2415e1c23e0aad8ca52358523fc9116a2eb6e4d4ef47577b1635c7cee3d5f'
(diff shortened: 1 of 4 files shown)


h0tw1r3:
one step closer to community being able to build foss packages without hacking up vanagon configs

h0tw1r3:
tldr; I think just patching out the automake version requirement will work for most (all?) platforms.

Did a little digging. I don't think the custom pre-autoreconf is necessary with the exception of aix 7.1.
The automake 1.15 requirement was added in libxml2 2.10, and it looks like it was arbitrary, moved from the python automake to the base. Then later changed to 1.16.3 for fixing python 3.10 detection.

joshcooper:
Thanks for looking into this @h0tw1r3! It turns out that `autoreconf -fi` ends up calling on `libtool` and we have some older platforms where that fails. I was able to instead switch to using the release libxml2 and libxslt tarballs (see #827) so I'm going to close this.

PR #770, opened 2023-11-28
(PA-5902) Enable puppet-runtime builds for macOS 14(ARM)

amitkarsale:
Enable puppet-runtime builds for macOS 14(ARM)


joshcooper:
We'll need to get a vanagon release out with this change https://github.com/puppetlabs/vanagon/pull/831

joshcooper (approved):

PR #773, opened 2023-12-05
(PA-5751) debian 12 arm platform support

h0tw1r3:
[vanagon-generic-builder agent-runtime-main](https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2549/)
[vanagon-generic-builder agent-runtime-7.x](https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2551/)

### DIFF
--- a/configs/platforms/debian-12-armhf.rb
+++ b/configs/platforms/debian-12-armhf.rb
@@ -1,0 +1,30 @@
+platform "debian-12-armhf" do |plat|
+  plat.servicedir "/lib/systemd/system"
+  plat.defaultdir "/etc/default"
+  plat.servicetype "systemd"
+  plat.codename "bullseye"
+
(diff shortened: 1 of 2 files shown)


configs/platforms/debian-12-armhf.rb
```diff
@@ -0,0 +1,30 @@
+platform "debian-12-armhf" do |plat|
```
amitkarsale, line 1: How is this file difference from `configs/platforms/debian-12-aarch64.rb`
Can you share why do we need this file as for previous debian platforms we dont have this file.

h0tw1r3: This is for the ARMv7 architecture (eg. raspberry pi?)

h0tw1r3: The platforms are also there for debian-10 and debian-11.

PR #774, opened 2023-12-05
 (PA-5550) debian 12 x86_64 platform support

h0tw1r3:
[vanagon-generic-builder agent-runtime-main](https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2550/)
[vanagon-generic-builder agent-runtime-7.x](https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2552/)

### DIFF
--- a/configs/components/_base-ruby-selinux.rb
+++ b/configs/components/_base-ruby-selinux.rb
@@ -21,6 +21,12 @@ elsif platform.name.start_with?('el-9')
   pkg.version '3.3'
   pkg.sha256sum 'acfdee27633d2496508c28727c3d41d3748076f66d42fccde2e6b9f3463a7057'
   pkg.url "https://github.com/SELinuxProject/selinux/releases/download/#{pkg.get_version}/libselinux-#{pkg.get_version}.tar.gz"
+elsif platform.name.start_with?('debian-12')
+  # SELinux 3.4 is the minimum version available in Debian 12 repos
+  pkg.version '3.4'
+  pkg.sha256sum '77c294a927e6795c2e98f74b5c3adde9c8839690e9255b767c5fca6acff9b779'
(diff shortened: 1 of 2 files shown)


shubhamshinde360 (approved):

configs/components/_base-ruby-selinux.rb
```diff
@@ -21,6 +21,12 @@
   pkg.version '3.3'
   pkg.sha256sum 'acfdee27633d2496508c28727c3d41d3748076f66d42fccde2e6b9f3463a7057'
   pkg.url "https://github.com/SELinuxProject/selinux/releases/download/#{pkg.get_version}/libselinux-#{pkg.get_version}.tar.gz"
+elsif platform.name.start_with?('debian-12')
```
amitkarsale, line 24: What made us add these conditions, in the past we have added platforms but never required a change here. Just want to enhance my knowledge.

h0tw1r3: @amitkarsale building against a much older libselinux will result in broken runtime bindings. [See PA-5632](https://perforce.atlassian.net/issues/PA-5632) for details on why the 3.3 was added. While we don't actively test selinux on Ubuntu or Debian (I would like to see that feature added in the future) this ensures the build works properly and built against the vendor shipped version.

As to why the vendor libselinux devel package isn't used instead of downloading the source, I'm not sure. It's mentioned in PA-5632 to keep using source.

h0tw1r3: @amitkarsale Can you link to the failed pipeline please? I don't see any failures.

PR #777, opened 2023-12-11
(PA-5929) Don't overwrite windows specific install blocks

tvpartytonight:
The placement of the logic to enable/disable the system-ffi should not overwrite the windows specific install blocks.


tvpartytonight:
I built `agent-runtime-main` successfully on  `windows-2012r2-x86`.

AriaXLi (approved):

PR #780, opened 2023-12-20
(PA-5960) Bump vendored puppet gems for concurrent-ruby 1.2.2 compat

joshcooper:
Commit f201ef9 bumped all runtime projects to use concurrent-ruby 1.2.2, but some runtime projects vendor an older version of puppet that requires concurrent-ruby less than 1.2. So bump those runtime projects to use the latest released puppet version.

Technically, puppet versions 7.25.0 and 8.1.0 were the first gems to support concurrent-ruby 1.2 and up, but there are some other bug fixes with binary Strings and monkey patching `File,Dir.exists?` back, so bump to latest.

### DIFF
--- a/configs/components/rubygem-puppet.rb
+++ b/configs/components/rubygem-puppet.rb
@@ -4,9 +4,21 @@ component 'rubygem-puppet' do |pkg, settings, platform|
   version = settings[:rubygem_puppet_version] || '6.28.0'
   pkg.version version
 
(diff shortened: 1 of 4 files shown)


joshcooper:
I manually built these on RHEL7:

* bolt-runtime
* pe-bolt-server-runtime-main
* pe-bolt-server-runtime-2021.7.x

mhashizume (approved):

PR #787, opened 2024-02-02
(PA-5999) Add support for SLES-11 (Intel)

shubhamshinde360:
- Removed the java package because its no longer needed and removed gcc(version 4.3.4 in the image) because we are supposed to be using pl-gcc (version 4.8.4).
 - Added SLES-11 to without-dtrace platforms to build ruby
 - Add --with-baseruby=no to configure script to not use base ruby since the image has ruby 1.8.7.

### DIFF
--- a/configs/platforms/sles-11-x86_64.rb
+++ b/configs/platforms/sles-11-x86_64.rb
@@ -8,8 +8,6 @@ platform "sles-11-x86_64" do |plat|
     "aaa_base",
     "autoconf",
     "automake",
-    "gcc",
-    "java-1_7_1-ibm-devel",
     "libbz2-devel",
     "make",
     "pkgconfig",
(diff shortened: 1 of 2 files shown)


configs/platforms/sles-11-x86_64.rb
```diff
@@ -19,6 +17,16 @@
...
+  plat.provision_with(%q{cat <<END > /etc/zypp/repos.d/localmirror-os.repo
+    [localmirror-os]
+    name=localmirror-os
+    baseurl=http://osmirror.delivery.puppetlabs.net/sles-11-sp4-x86_64/RPMS.os
+    enabled=1
+    gpgcheck=0
+    gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-SLES-11
+    autorefresh=1
+    type=rpm-md 
+    END})
```
joshcooper, line 29: Pasting my note from slack, this unfortunately didn't work for me as the rest of vanagon's provision commands were written into the `localmirror-os.repo` file

shubhamshinde360:
Resolved under https://github.com/puppetlabs/puppet-runtime/pull/788

PR #792, opened 2024-02-20
(PA-5999) Use pl-autotools instead of autoconf and automake for SLES-11

shubhamshinde360:
- For Puppet8, we use ruby3 which requires autoconf to be >= 2.67, but SLES-11 supports only upto 2.63. So, we use pl-autotools for both puppet7 and puppet8 for consistency.


amitkarsale:
I guess we have a dependency of autoconf(or autoreconf). Omitting it is failing for `autoreconf`.

joshcooper (approved):

PR #796, opened 2024-02-26
(maint) Use Nori 2.6.0 for pe-installer-runtime-2021.7.x

nmburgan:
2.7.0 requires Ruby 3, so include this at 2.6 so Bolt doesn't try pulling in 2.7.


donoghuc (approved):

donoghuc:
we should really just do https://perforce.atlassian.net/browse/PE-36254

PR #800, opened 2024-03-04
(PA-6025) add ubuntu 24.04 amd64 platform

h0tw1r3:


h0tw1r3:
https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/1433/

configs/projects/bolt-runtime.rb
```diff
@@ -230,7 +230,7 @@
 
   # Components from puppet-runtime included to support apply on localhost
   # Only bundle SELinux gem for RHEL,Centos,Fedora
-  if platform.is_el? || platform.is_fedora?
+  if platform.is_el? || platform.is_fedora? || platform.is_debian?
```
joshcooper, line 233: Will this break bolt on Debian 10 - 12 or Ubuntu < 24.04?

h0tw1r3: It shouldn't break as puppet-runtime includes selinux gem support for ubuntu/debian,, but it should be a separate PR.
Removed bolt related changes.

joshcooper (approved):

PR #802, opened 2024-03-11
(PA-5998) Add sles11-x86_64 to puppet-runtime

amitkarsale:
Enforcing runtime build to use latest gcc i.e `pl-gcc8`


configs/platforms/sles-11-x86_64.rb
```diff
@@ -4,17 +4,17 @@
...
-    "rsync",
-    "zlib-devel"
+  packages = %w[
+    aaa_base
+    libbz2-devel
+    make
+    pkgconfig
+    pl-autotools
+    pl-cmake
+    pl-gcc8
```
joshcooper, line 14: Since this platform definition is shared by both agent-runtime-7.x and agent-runtime-main, I'm thinking it would be better to install `pl-gcc8` in https://github.com/puppetlabs/puppet-runtime/blob/531fb1fd6b6740d793669f0514ee54b8cb027d19/configs/components/runtime-agent.rb#L6-L10

joshcooper (approved):

PR #803, opened 2024-03-14
(PE-37599) Build client-tools-runtime-2021.7.x project for Red Hat 9 …

imaqsood:
…FIPS (Intel)


skyamgarp (approved):

imaqsood:
> I see what's going on now. I think we should use continue using openssl 1.1.1k, so change this:
> 
> ```ruby
> elsif platform.name =~ /^redhatfips-[89]/
>   proj.setting(:openssl_version, '1.1.1k')
> ```
> 
> You also need to do the following below on line 114, so that we build our openssl-1.1.1k component:
> 
> ```ruby
> elsif (platform.name =~ /^redhatfips-[89]/ && proj.openssl_version == '1.1.1k'
>   proj.component "openssl-1.1.1-fips"
> ```
> 
> For example see how the agent does the same:
> 
> https://github.com/puppetlabs/puppet-runtime/blob/096e683a7bbff31f339799f3531af79d95ef50fd/configs/projects/_shared-agent-components.rb#L26-L27

I tested this PR with OpenSSL 1.1.1k before, but we couldn't install it during the build. The error said it couldn't find a match. I also tried with OpenSSL-1.1.1, but got the same error.

joshcooper:
For posterity, this changes RHELFIPS 8 and 9:

```
$ bundle exec rake vanagon:component_diff -- -P client-tools-runtime-2021.7.x -p all --from upstream/master --to HEAD 
...
Here is what your code changes would affect:

Project client-tools-runtime-2021.7.x

Platform name: redhatfips-8-x86_64
    Component 'curl'
        Field: build_requires[0]
        --------------------
        - openssl-1.1.1k
        + openssl-1.1.1-fips


    Component 'puppet-ca-bundle'
        Field: build_requires[0]
        --------------------
        - openssl-1.1.1k
        + openssl-1.1.1-fips


Platform name: redhatfips-9-x86_64
    Component 'openssl-1.1.1' was removed, not showing diff for it

    Component 'openssl-1.1.1-fips' was newly added, not showing diff for it

    Component 'curl'
        Field: build_requires[0]
        --------------------
        - openssl-1.1.1
        + openssl-1.1.1-fips


    Component 'puppet-ca-bundle'
        Field: build_requires[0]
        --------------------
        - openssl-1.1.1
        + openssl-1.1.1-fips
```

configs/projects/_shared-client-tools-runtime.rb
```diff
@@ -25,8 +25,8 @@
 # and krb5-devel
 if proj.settings[:openssl_version]
   # already defined in the project
-elsif platform.name =~ /^redhatfips-8/
-  proj.setting(:openssl_version, '1.1.1k')
+elsif platform.name =~ /^redhatfips-[89]/
```
joshcooper, line 28: can you update the `define 1.1.1k here` comment above?

joshcooper:
It looks like `configs/projects/pe-installer-runtime-2021.7.x.rb` has the same problem, but since we don't support RHEL9 FIPS primary servers yet, then it's not an issue we need to solve right now. But perhaps file a JIRA ticket so we remember to fix it later?

```
$ git grep 1.1.1k configs/projects/
configs/projects/_shared-client-tools-runtime.rb:# We need to explicitly define 1.1.1k here to avoid
configs/projects/pe-installer-runtime-2021.7.x.rb:  # We need to explicitly define 1.1.1k here to avoid
configs/projects/pe-installer-runtime-2021.7.x.rb:    proj.setting(:openssl_version, '1.1.1k')
```

joshcooper (approved):

PR #804, opened 2024-03-14
Apply Ruby patch to enable FIPS with OpenSSL 3

mhashizume:
Prior to this commit, attempting to enable fips_mode in Ruby 3.2.3's OpenSSL gem would cause the error "This version of OpenSSL does not support FIPS mode (OpenSSL::OpenSSLError)". This was due to OpenSSL removing the FIPS_mode() function in OpenSSL 3, which Ruby previously relied on to detect if FIPS was enabled.

Upstream has fixed this incompatibility with OpenSSL 3 in Ruby 3.3. The contributor has indicated that the fix should be backported to Rubies 3.0-3.2 in the future.

This commit applies Ruby's fix as a patch.

See also:
- ruby/ruby@678d41bc51fe31834eec0b653ba0e47de5420aa0
- https://bugs.ruby-lang.org/issues/20000

### DIFF
--- a/configs/components/ruby-3.2.3.rb
+++ b/configs/components/ruby-3.2.3.rb
@@ -55,6 +55,12 @@ component 'ruby-3.2.3' do |pkg, settings, platform|
     pkg.apply_patch "#{base}/revert_ruby_utf8_default_encoding.patch"
   end
 
+  if platform.is_fips?
+    # This is needed on Ruby < 3.3 until the fix is backported (if ever)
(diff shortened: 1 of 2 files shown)


mhashizume:
The original Ruby patch modified two files, but I removed the patched test file as it caused compilation failures that I did not fully understand (and I assumed that we don't need to ship Ruby's tests).

joshcooper (approved):

PR #806, opened 2024-03-22
(PE-36494,PE-37833) Standardize pe-installer-runtime on shared format with more deps

nmburgan:
This standardizes the pe-installer-runtime definition to the format used by the other runtimes, using a shared definition for common things between 2021.7.x and main. It also expands the list of components so we are providing all dependencies to Bolt and Puppet. This is so that we don't get surprised when pe-installer-vanagon gem installs Bolt and pulls in dependencies that may break on the version of Ruby we're using.

### DIFF
--- a/configs/projects/_shared-pe-installer-runtime.rb
+++ b/configs/projects/_shared-pe-installer-runtime.rb
@@ -1,0 +1,156 @@
+platform = proj.get_platform
+proj.version_from_git
+proj.generate_archives true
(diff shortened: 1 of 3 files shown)


donoghuc (commented):
Do you want to just manage bolt as a gem here too? Unless i'm missing something I dont think you would ever need to do a `gem install bolt` for installer.

donoghuc (approved):

PR #808, opened 2024-03-26
(maint) Add net-ssh version back into installer runtime 2021.7.x

nmburgan:
Accidentally removed this from the new pe-installer-runtime-2021.7.x, but since main and 2021.7.x use different versions this needs to stay in here.


jpartlow (approved):
I see, main sets 7.0.1, but 2021.7.x wasn't setting anything and ending up with a default net-ssh that was out of date for our needs.

PR #809, opened 2024-03-27
(PE-36494) Ship puppet gem with installer runtime

donoghuc:


donoghuc:
I'll ticket out bumping gem deps (including puppet) after platform release.

PR #817, opened 2024-04-04
Drop unused win32 gems from bolt-runtime and other cleanup

joshcooper:
Puppet 6 was the last stream to depend on win32-* gems, see https://github.com/puppetlabs/puppet-runtime/commit/513c5583bb6449187eb7ef6306997398917a5c10. In
Puppet 7, all of the required functionality was implemented in puppet itself.

Since bolt-runtime has moved to puppet 7.27, we can drop these gems.


donoghuc (approved):
Nice, i didnt know this information.

PR #818, opened 2024-04-05
Load cert related files in binary mode on Windows

joshcooper:
See patch for details.


joshcooper:
There's a better way to do this, closing for now

PR #832, opened 2024-04-19
(maint) Use packaged libxml to build nokogiri on MacOS

cthorn42:
This PR changes the gem install command for the Nokogiri gem. Currently when building nokogiri it uses the system libxml2 and libxslt. We now will pass in an option in the `gem install` command to use the expected libraries.

### DIFF
--- a/configs/components/_base-rubygem.rb
+++ b/configs/components/_base-rubygem.rb
@@ -37,6 +37,15 @@ end
 pkg.url("https://rubygems.org/downloads/#{name}-#{version}.gem")
 pkg.mirror("#{settings[:buildsources_url]}/#{name}-#{version}.gem")
 
-pkg.install do
-  "#{settings[:gem_install]} #{name}-#{version}.gem"
+# If a gem needs more command line options to install set the :gem_install_options
+# in its component file rubygem-<compoment>, before the instance_eval of this file.
+if settings[:gem_install_options].nil?
+  pkg.install do
+    "#{settings[:gem_install]} #{name}-#{version}.gem"
+  end
+else
+  pkg.install do
+    "#{settings[:gem_install]} #{name}-#{version}.gem #{settings[:gem_install_options]}"
+  end
 end
+
--- a/configs/components/rubygem-nokogiri.rb
+++ b/configs/components/rubygem-nokogiri.rb
@@ -1,18 +1,24 @@
 component 'rubygem-nokogiri' do |pkg, _settings, _platform|
   pkg.version '1.14.2'
   pkg.sha256sum 'c765a74aac6cf430a710bb0b6038b8ee11f177393cd6ae8dadc7a44a6e2658b6'
+  # On macOS when we are not cross compiling we need to use runtime's libxml2 and libxslt
+  if platform.is_macos? && !platform.is_cross_compiled?
+    settings[:gem_install_options] = "-- --use-system-libraries \
+        --with-xml2-lib=#{settings[:libdir]} \
+        --with-xml2-include=#{settings[:includedir]}/libxml2 \
+        --with-xslt-lib=#{settings[:libdir]} \
+        --with-xslt-include=#{settings[:includedir]}"
+  end
   instance_eval File.read('configs/components/_base-rubygem.rb')
-
   pkg.build_requires 'rubygem-mini_portile2'
-
   gem_home = settings[:gem_home]
   pkg.environment "GEM_HOME", gem_home
-
-  # When cross compiling nokogiri native extensions on macOS 11/12 ARM, there is a 94M tmp
-  # directory that's not needed
-  if platform.is_macos? && platform.architecture == 'arm64'
-    install do
-      "rm -r #{gem_home}/gems/nokogiri-#{pkg.get_version}/ext/nokogiri/tmp"
+  if platform.is_macos?
+    pkg.environment "PKG_CONFIG_PATH", "#{settings[:libdir]}/pkgconfig"
+    if platform.is_cross_compiled?
+      pkg.install do
+        "rm -r #{gem_home}/gems/nokogiri-#{pkg.get_version}/ext/nokogiri/tmp"
+      end
     end
   end
 end


configs/components/_base-rubygem.rb
```diff
@@ -37,6 +37,8 @@
...
 pkg.mirror("#{settings[:buildsources_url]}/#{name}-#{version}.gem")
 
-pkg.install do
-  "#{settings[:gem_install]} #{name}-#{version}.gem"
-end
+unless name =~ /nokogiri/
+  pkg.install do
+    "#{settings[:gem_install]} #{name}-#{version}.gem"
+  end
+end
```
cthorn42, line 44: This is just a placeholder, I think the long term solution is to change this to just add to the install command something like `#{pkg.gem_install_options}`, where we define the pkg.gem_install_options in the nokogiri component.
I can't quite wrap my head around how to implement this yet though.

configs/components/rubygem-nokogiri.rb
```diff
@@ -7,12 +7,15 @@
...
-
-  # When cross compiling nokogiri native extensions on macOS 11/12 ARM, there is a 94M tmp
-  # directory that's not needed
-  if platform.is_macos? && platform.architecture == 'arm64'
-    install do
-      "rm -r #{gem_home}/gems/nokogiri-#{pkg.get_version}/ext/nokogiri/tmp"
+  if platform.is_macos?
+    pkg.environment "PKG_CONFIG_PATH", "#{settings[:libdir]}/pkgconfig"
+    pkg.install do
+      "#{settings[:gem_install]} nokogiri --platform=ruby -- --use-system-libraries --with-opt-dir=/opt/puppetlabs/puppet"
```
joshcooper, line 13: One subtle issue, I think `gem install nokgiri` behaves differently than `gem install nokogiri.gem`. I think the former will download the gem from rubygems/artifactory, while the latter will see that it's a local gem on disk and should install that.

cthorn42: I tried this, specifying `#{name}-#{pkg.get_version}.gem`, and got the same warning about libxml: `WARNING: Nokogiri was built against libxml version 2.9.13, but has dynamically loaded 2.12.6`.

cthorn42:
Here is my current work in progress for runtime, the otool output for the nokogiri bundle:
```
otool -L /opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/lib/nokogiri/nokogiri.bundle
/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/lib/nokogiri/nokogiri.bundle:
	/opt/puppetlabs/puppet/lib/libruby.3.2.dylib (compatibility version 3.2.0, current version 3.2.3)
	/opt/puppetlabs/puppet/lib/libexslt.0.dylib (compatibility version 9.0.0, current version 9.21.0)
	/opt/puppetlabs/puppet/lib/libxslt.1.dylib (compatibility version 3.0.0, current version 3.39.0)
	/opt/puppetlabs/puppet/lib/libxml2.2.dylib (compatibility version 15.0.0, current version 15.6.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```
This looks right when compared to the same output on the currenty nightly osx-13-arm64 output:
```
otool -L /opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/lib/nokogiri/nokogiri.bundle
/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/lib/nokogiri/nokogiri.bundle:
	/opt/puppetlabs/puppet/lib/libruby.3.2.dylib (compatibility version 3.2.0, current version 3.2.3)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/liblzma.5.dylib (compatibility version 6.0.0, current version 6.3.0)
```
Note in my work in progress we see the libxml2 and libxslt are using the bundled libraries in my work in progress, and in comparission the current nightlies as expected does not load our bundled libraries.

Now where things look wrong is when running the Nokogiri binary, this is how it the binary looks when run from the current nightly build:
```
 /opt/puppetlabs/puppet/bin/nokogiri --version
# Nokogiri (1.14.2)
    ---
    warnings: []
    nokogiri:
      version: 1.14.2
      cppflags:
      - "-I/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri"
      - "-I/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri/include"
      - "-I/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri/include/libxml2"
      ldflags: []
    ruby:
      version: 3.2.3
      platform: arm64-darwin22
      gem_platform: arm64-darwin-22
      description: ruby 3.2.3 (2024-01-18 revision 52bb2ac0a6) [arm64-darwin22]
      engine: ruby
    libxml:
      source: packaged
      precompiled: false
      patches:
      - 0001-Remove-script-macro-support.patch
      - 0002-Update-entities-to-remove-handling-of-ssi.patch
      - 0003-libxml2.la-is-in-top_builddir.patch
      - '0009-allow-wildcard-namespaces.patch'
      libxml2_path: "/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri"
      memory_management: ruby
      iconv_enabled: true
      compiled: 2.10.3
      loaded: 2.10.3
    libxslt:
      source: packaged
      precompiled: false
      patches:
      - 0001-update-automake-files-for-arm64.patch
      datetime_enabled: true
      compiled: 1.1.37
      loaded: 1.1.37
    other_libraries:
      libgumbo: 1.0.0-nokogiri
```
Things to note in the current nightly build above, the vendored libxml is used but the system libxslt is used. The goal of this ticket is to get the libxslt to use our libxslt.
When we look at my current work in progress Nokogiri binary output we see some issues:
```
./puppet/bin/nokogiri --version
WARNING: Nokogiri was built against libxml version 2.9.13, but has dynamically loaded 2.12.6
# Nokogiri (1.14.2)
    ---
    warnings:
    - Nokogiri was built against libxml version 2.9.13, but has dynamically loaded 2.12.6
    nokogiri:
      version: 1.14.2
      cppflags:
      - "-I/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri"
      ldflags: []
    ruby:
      version: 3.2.3
      platform: arm64-darwin22
      gem_platform: arm64-darwin-22
      description: ruby 3.2.3 (2024-01-18 revision 52bb2ac0a6) [arm64-darwin22]
      engine: ruby
    libxml:
      source: system
      memory_management: ruby
      iconv_enabled: false
      compiled: 2.9.13
      loaded: 2.12.6
    libxslt:
      source: system
      datetime_enabled: true
      compiled: 1.1.39
      loaded: 1.1.39
    other_libraries:
      libgumbo: 1.0.0-nokogiri
```
First there is an obvious warning message about libxml, then we see that the system libxml and system libxslt are being used. That seems to be a problem.

joshcooper:
I think the nightly version is correctly using the libxml2 headers that are vendored in the nokogiri gem as can be seen by this entry in `cppflags`:

```
      - "-I/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2/ext/nokogiri/include/libxml2"
```

But the work is progress seems to be using `compiled: 2.9.13` which is not the version in the nokogiri gem and it's not what we're using in the runtime. So I'm guessing the build is using `/usr/include`.

You can go full explicit mode and list the include and lib directories for each component like:

```
gem install nokogiri -- \
    --use-system-libraries \
    --with-xml2-lib=/path/to/builds/lib \
    --with-xml2-include=/path/to/builds/include/libxml2 \
    --with-xslt-lib=/path/to/builds/lib \
    --with-xslt-include=/path/to/builds/include
```

Can also use `settings[:includedir]` and settings[:libdir]` instead of hardcoding `/opt/puppetlabs/puppet/include`, etc

configs/components/_base-rubygem.rb
cthorn42: Thanks Josh, took a bit to wrap my head around things but the prior to the instance_eval was the key point. I've pushed up my changes. Working on last little bit now, vanagon-generic-job is failing osx-12-arm, which is cross compiled. 
Thats the last thing I need to resolve.

cthorn42:
I've removed the block label, validation for this is complete. Here is the agent-runtime-main vanagon generic job link: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2907/ (ignore the one failure, that is an AWS transient)
There is one issue though, the gem that gets installed is x86_64 and not arm. 
So what I'm suggesting is we leave Nokogiri on MacOS support for version 11/12 ARM to be as it is; broken. 
We have it fixed going forward starting with MacOS 13 and thats good enough.

mhashizume (approved):

PR #837, opened 2024-04-24
(CAT-1789) Add Windows 11 platform definition

david22swan:
Required for the PDK to support Windows 11
To start with I have imitated the settings used with the windows server 2019.


joshcooper:
Hi @david22swan generally speaking, if PDK is being built on the oldest supported Windows OS, then it should install and run on all newer versions. For example, we build the agent runtime on 2012R2, but install/test on all other Windows platforms. Can't PDK do the same?

david22swan:
@joshcooper Thinking on it, that would probably be a better idea.
Will look into and and see about testing it out.
Will close this for now, thanks for the suggestion

PR #845, opened 2024-05-08
fix augeas pkg.url

h0tw1r3:
latest versions are not available on download.augeas.net

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -15,7 +15,13 @@ component 'augeas' do |pkg, settings, platform|
     raise "augeas version #{version} has not been configured; Cannot continue."
   end
 
-  pkg.url "http://download.augeas.net/augeas-#{pkg.get_version}.tar.gz"
+  # releases from 1.13.0 onward are only available from github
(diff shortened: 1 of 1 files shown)


configs/components/augeas.rb
```diff
@@ -15,7 +15,7 @@
     raise "augeas version #{version} has not been configured; Cannot continue."
   end
 
-  pkg.url "http://download.augeas.net/augeas-#{pkg.get_version}.tar.gz"
+  pkg.url "https://github.com/hercules-team/augeas/releases/download/release-#{pkg.get_version}/augeas-#{pkg.get_version}.tar.gz"
```
joshcooper, line 18: Looks like this broke in 9c6456a35f3f044d3973e2e8dc341dee8eb8a72e AFAIK, only 1.13.0+ is on github

h0tw1r3: @joshcooper fixed!

joshcooper (approved):
not blocking but maybe add a comment that the fallback to download.augeas.net can be removed once we stop supporting augeas 1.12.0

PR #853, opened 2024-05-21
(CAT-1871) Add `strscan` as a PDK dependency for windows.

david22swan:
The PDK Build step for windows is currently hitting an error related to `strscan`


david22swan:
Unfortunately does not seem to fix the issue.

joshcooper:
strscan is a default gem (so it ships with ruby) so in theory this shouldn't be needed. However, there is bug in rexml 3.2.7 and 3.2.8, which requires a newer version of strscan than ships in some ruby versions. You can temporarily pin back to rexml 3.2.6 or wait for the 3.2.9 to come out

david22swan:
> strscan is a default gem (so it ships with ruby) so in theory this shouldn't be needed. However, there is bug in rexml 3.2.7 and 3.2.8, which requires a newer version of strscan than ships in some ruby versions. You can temporarily pin back to rexml 3.2.6 or wait for the 3.2.9 to come out

Thx for the help, will take a look into this

PR #854, opened 2024-05-21
(maint) Bump bcrypt gem

donoghuc:


donoghuc:
So if i understand correctly @h0tw1r3 we dont need to explicitly vendor mingw gems https://rubygems.org/gems/bcrypt_pbkdf/versions/1.1.1 because we do that as a build step? We do that in some cases for ffi, but that is a whole other can of worms.

h0tw1r3:
> So if i understand correctly @h0tw1r3 we dont need to explicitly vendor mingw gems https://rubygems.org/gems/bcrypt_pbkdf/versions/1.1.1 because we do that as a build step? We do that in some cases for ffi, but that is a whole other can of worms.

Correct. The only difference between the 1.1.0 release and 1.1.1 is it replaces GNU specific things to compile correctly on all platforms and compilers:

* https://github.com/net-ssh/bcrypt_pbkdf-ruby/pull/22/commits/522e0baef6ab70ffdba1a68c85749b49be366f88 - fix compile on windows missing bzero 
   bzero is not standard C but POSIX, and deprecated in POSIX.1-2001 and removed in POSIX.1-2008.
* https://github.com/net-ssh/bcrypt_pbkdf-ruby/pull/22/commits/ee964c8847b7f4867da8052f041bbc44025d1cf5 - replace strings include with string for compiler compatibilty

joshcooper (approved):

donoghuc:
Thanks!

PR #855, opened 2024-05-23
Remove base64 and nkf which conflict with default gems

tlehman:
# Testing

I built this and confirmed there are no duplicates of these standard gems:
```
[root@perky-artillery gems]# ls | grep nkf
nkf-0.1.2
[root@perky-artillery gems]# ls | grep base
base64-0.1.1
```


joshcooper:
base64 is a default gem for ruby 3.2.2 https://stdgems.org/3.2.2/#bundled-gems 

But `nkf` is not. It be good to understand why it was added as a dependency and what might break if we remove it. I'm guessing it was for CFPropertyList based on https://rubygems.org/gems/nkf/reverse_dependencies?

tlehman:
@joshcooper I just checked https://stdgems.org/nkf/ and it's a default gem. I also just built this branch and verified that it includes nkf 0.1.2 (only), so this branch is ready to merge.

donoghuc (approved):
CI should be green now. I think this is g2g.

joshcooper (approved):

PR #860, opened 2024-06-12
(PA-6549) Fix warning:undefining the allocator of T_DATA class

AriaXLi:
After puppetlabs/puppet#9349 was merged the
       `warning: undefining the allocator of T_DATA class SWIG::TYPE_p_selabel_handle`
error popped up again on RHEL-8 x86_64.

We ran into this previously and thought puppetlabs/puppet-runtime#620 fixed it completely. This issue was completely fixed in swig 4.2.0+.

This commit patches selinuxswig_ruby_wrap.c with swig/swig@aa21014 and is applied after the patch from puppetlabs/puppet-runtime#620 since the fix in swig/swig@aa21014 came after.

### DIFF
--- a/configs/components/_base-ruby-selinux.rb
+++ b/configs/components/_base-ruby-selinux.rb
@@ -4,6 +4,7 @@
 #
 
 pkg.add_source("file://resources/patches/ruby-selinux/selinuxswig_ruby_wrap.patch")
+pkg.add_source("file://resources/patches/ruby-selinux/selinuxswig_ruby_undefining_allocator.patch")
 
 # These can be overridden by the including component.
 ruby_version ||= settings[:ruby_version]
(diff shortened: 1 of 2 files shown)


AriaXLi:
Successfully built in vanagon-generic-builder with `el-9-x86_64 el-8-x86_64 debian-11-aarch64 debian-12-amd64` as `BUILD_TARGETS` to ensure the existing patch is correctly applied or only applied for platforms that use Ruby 3+: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/1578/

AriaXLi:
I also successfully built locally:

```
$ bundle exec vanagon build agent-runtime-main el-8-x86_64 --preserve always
...
Warning: Permanently added 'very-stint.delivery.puppetlabs.net' (ED25519) to the list of known hosts.

[root@very-stint libselinux-2.9]# /opt/puppetlabs/puppet/bin/ruby -e 'puts require "selinux"; Selinux.selabel_open(0, nil, 0)'
true
```
I also confirmed that the patch is correctly applied for RHEL8

joshcooper (approved):

PR #868, opened 2024-07-10
(PA-6507) Patch rexml gem for CVE-2024-35176

shubhamshinde360:
- Ruby 3.2.4 has rexml as its bundled gem, so the oatch was applied after the install step since bundled gems are available in the build after the install step in the .bundle folder.
 - Ruby 2.7.8 has rexml as its default gem, so we can go with the usual way to patch it.
 - Note that in the patch files for ruby_32, the file paths are prefixed with .bundle/gems/rexml-3.2.5 because that is where the bundled gem rexml is located with respect to the working directory of patching.
 - The upstream fix commit: https://github.com/ruby/rexml/commit/4325835f92f3f142ebd91a3fdba4e1f1ab7f1cfb

### DIFF
--- a/resources/patches/ruby_27/rexml_for_CVE-2024-35176.patch
+++ b/resources/patches/ruby_27/rexml_for_CVE-2024-35176.patch
@@ -1,0 +1,135 @@
+commit e597f07718c27dc4414ad39f121376e53056475a
+Author: Shubham Shinde <shubham.shinde@perforce.com>
+Date:   Tue Jul 9 19:40:43 2024 +0530
+
+    Read quoted attributes in chunks (#126)
+    
+    
+    # Conflicts:
+    #       lib/rexml/parsers/baseparser.rb
+    #       lib/rexml/source.rb
+
+diff --git a/lib/rexml/parsers/baseparser.rb b/lib/rexml/parsers/baseparser.rb
+index f76aed0..f0b365d 100644
+--- a/lib/rexml/parsers/baseparser.rb
++++ b/lib/rexml/parsers/baseparser.rb
+@@ -518,25 +518,43 @@ module REXML
+               message = "Missing attribute equal: <#{name}>"
+               raise REXML::ParseException.new(message, @source)
+             end
+-            quote = scanner.scan(/['"]/)
+-            unless quote
++            unless match = @source.match(/(['"])/, true)
+               message = "Missing attribute value start quote: <#{name}>"
+               raise REXML::ParseException.new(message, @source)
+             end
+-            unless scanner.scan(/.*#{Regexp.escape(quote)}/um)
+-              match_data = @source.match(/^(.*?)(\/)?>/um, true)
+-              if match_data
+-                scanner << "/" if closed
+-                scanner << ">"
+-                scanner << match_data[1]
+-                scanner.pos = pos
+-                closed = !match_data[2].nil?
+-                next
+-              end
+-              message =
+-                "Missing attribute value end quote: <#{name}>: <#{quote}>"
++            quote = match[1]
++            value = @source.read_until(quote)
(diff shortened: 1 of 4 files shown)


shubhamshinde360:
Tested on vanagon generic build
agent-runtime-7.x:
https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3039/ 
(Please ignore aix-7.2-ppc as we don't build it for 7.x but I had accidentaly added it)

agent-runtime-main:
https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3040/
Seems like there is some generic issue with solaris-11-sparc since it is also failing for the puppetlabs/master branch with the same error as above.

Artifacts: https://builds.delivery.puppetlabs.net/puppet-runtime/e4e77d3944647462fea2119e1080940d50e5db81/artifacts

joshcooper:
> Seems like there is some generic issue with solaris-11-sparc

You'll need to use `solaris-11-native-sparc` when building `agent-runtime-main`. The `solaris-11-sparc` platform is for `agent-runtime-7.x` and is cross-compiled.

To make sure you're using the correct platforms, I recommend copying the `BUILD_TARGETS` from the appropriate init job:

[agent-runtime-7.x](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-runtime/view/agent-runtime-7.x/job/platform_agent-runtime_runtime-vanagon-init_7.x/)
[agent-runtime-main](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-runtime/view/agent-runtime-main/job/platform_agent-runtime_runtime-vanagon-init_main/)

mhashizume:
Closing and re-opening to rekick checks.

joshcooper:
@shubhamshinde360 I noticed the vanagon generic builder jobs only included a subset of platforms, e.g. included osx-14-arm64, but not osx-12-arm64. Because the former is native compiled, but the latter is cross compiled, we don't really know if this will compile on all platforms. Could you run against the full set for 7.x and main?

resources/patches/ruby_27/rexml_for_CVE-2024-35176.patch
```diff
@@ -0,0 +1,135 @@
+commit e597f07718c27dc4414ad39f121376e53056475a
+Author: Shubham Shinde <shubham.shinde@perforce.com>
+Date:   Tue Jul 9 19:40:43 2024 +0530
+
+    Read quoted attributes in chunks (#126)
+    
+    
+    # Conflicts:
+    #       lib/rexml/parsers/baseparser.rb
+    #       lib/rexml/source.rb
```
joshcooper, line 10: I wouldn't expect conflicts. Did this occur because you cherry-picked the commit into the ruby 2.7 source tree? Is there a different patch we can use that applies cleanly?

shubhamshinde360: Hi @joshcooper,
Yes, I cherry-picked the commit https://github.com/ruby/rexml/commit/4325835f92f3f142ebd91a3fdba4e1f1ab7f1cfb into tag https://github.com/ruby/ruby/releases/tag/v2_7_8 and it conflicted. 
Was not able to find other patches to apply cleanly either.

shubhamshinde360: We had to resolve conflicts manually, is there any other path we can take in such situations?

shubhamshinde360:
Superseded by https://github.com/puppetlabs/puppet-runtime/pull/873

PR #869, opened 2024-07-10
(CAT-1869) - Add puppet-modulebuilder to pdk bundled gems

jordanbreen28:
This PR adds the puppet-modulebuilder gem to the gems bundled with the pdk. This gem is now required for the module build functionality.

### DIFF
--- a/configs/projects/_pdk-components.rb
+++ b/configs/projects/_pdk-components.rb
@@ -91,6 +91,9 @@ proj.component 'rubygem-public_suffix'
 proj.component 'rubygem-addressable'
 proj.component 'rubygem-json-schema'
 
+# PDK build
+proj.component 'rubygem-puppet-modulebuilder'
+
 # Other deps
 proj.component 'rubygem-deep_merge'
 proj.component 'rubygem-json_pure'
--- a/configs/components/rubygem-puppet-modulebuilder.rb
+++ b/configs/components/rubygem-puppet-modulebuilder.rb
@@ -1,0 +1,6 @@
+component 'rubygem-puppet-modulebuilder' do |pkg, settings, platform|
+  pkg.version '1.0.0'
+  pkg.sha256sum 'cf9d9e8146aeae780b7c61f30847a4cb631debcf708c21281976d5ed79820cfd'
+
+  instance_eval File.read('configs/components/_base-rubygem.rb')
+end


jordanbreen28:
failures aren't something I've seen before.. Have verified they occur when my changes are backed out. Possibly transient?

configs/projects/_pdk-components.rb
```diff
@@ -91,6 +91,9 @@
 proj.component 'rubygem-addressable'
 proj.component 'rubygem-json-schema'
 
+# PDK build
+proj.component 'rubygem-puppet-modulebuilder'
+
```
joshcooper, line 96: One potential gotcha when adding a new gem to any runtime project is making sure its runtime dependencies are added too, and that the versions of those dependencies satisfy the constraints of the newly added gem. Looking at https://rubygems.org/gems/puppet-modulebuilder/versions/1.0.0 I see it depends on `minitar` and `pathspec`, both of which are already vanagon components. Could you confirm that the versions of the `rubygem-{minitar,pathspec}` components will satisfy https://github.com/puppetlabs/puppet-modulebuilder/blob/v1.0.0/puppet-modulebuilder.gemspec#L27-L28?

jordanbreen28: hey @joshcooper - thanks for that, I wasn't aware of this! After looking it seems both the minitar and pathspec gem component in vanagon satisfy the version requirements set in puppet-modulebuilder's .gemspec.

minitar vanagon gem version: `0.9`
modulebuilder `.gemspec`: `~> 0.9`

pathspec vanagon gem version: `1.1.3`
modulebuilder `.gemspec`: `'>= 0.2.1', '< 2.0.0'`

david22swan (approved):
LGTM

joshcooper (approved):

joshcooper:
Sorry forgot one thing, could you make sure you can build the pdk runtime on all supported platforms using Vanagon Generic Builder before merging? The reason is each commit to the runtime causes all runtimes to be rebuilt

jordanbreen28:
Thanks @joshcooper! Going off the other PRs to this repo, here's my best attempt at that..
https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3048/

PR #870, opened 2024-07-10
Switch to Ruby 3.3 for GitHub Actions

mhashizume:
This commit updates Ruby from 3.2 to 3.3 in our GitHub Actions. The latest Ubuntu 22.04 runner image from GitHub has started causing errors in our builds, possibly due to the new inclusion of Ruby 3.2 in the base image.


mhashizume:
It seems like this solves the issue, I'll report an issue to upstream about this.

cthorn42 (approved):

PR #875, opened 2024-07-16
(PA-6507) Cleanup bundled version of rexml (3.2.5) from ruby 3

shubhamshinde360:
- rexml is a bundled gem in ruby 3.
 - When we gem install rexml version 3.2.9 to resolve CVE 2024-35176, we end up having two versions of rexml.
 - rexml 3.2.5 which is shipped with ruby as its bundled gem and rexml 3.2.9 which we manually installed.
 - This causes 'Gem::Specification.reset:rexml' warning to go to stderr each time puppet runs.
 - Run 'gem cleanup rexml' so that it removes the 3.2.5 version.


shubhamshinde360:
This will need to be reworked in a different PR.

PR #876, opened 2024-07-17
(CAT-1903) Add support for RedHat 8 and 9 AARCH architecture to the PDK

david22swan:
- Update the git component so that a line rejecting `dh-autoreconf` as a dependency when building for RedHat 8 and 9 will do the same when building against their AARCH architecture versions.


jordanbreen28 (approved):
LGTM

PR #882, opened 2024-07-29
(PA-6507)(PA-6736) Gem install rexml to 3.3.2 for CVE-2024-35176 and …

shubhamshinde360:
…CVE-2024-39908 in 7.x

 - The CVEs were fixed from rexml version 3.2.7 (CVE-2024-35176) and 3.3.2 (CVE-2024-39908).
 - Patching for the CVE wasn't getting applied cleanly and had a lot of conflicts. So updated the gem version to 3.3.2 in the rexml component file.
 - This addresses the CVEs for 7.x.
 - For solaris-10-sparc and solaris-11-sparc, we ignore dependency when installing rexml since the ruby in these platforms tries to install strscan (rexml's dependency) but fails while building native extensions. We can ignore installing strscan since it is shipped with ruby 2.7.8 as its default gem.

### DIFF
--- a/configs/components/rubygem-rexml.rb
+++ b/configs/components/rubygem-rexml.rb
@@ -1,6 +1,14 @@
 component 'rubygem-rexml' do |pkg, settings, platform|
-  pkg.version '3.2.6'
-  pkg.md5sum 'a57288ae5afed07dd08c9f1302da7b25'
+  pkg.version '3.3.2'
+  pkg.md5sum '55d213401f5e6a7a83ff3d2cd64a23fe'
 
+  # If the platform is solaris with sparc architecture in agent-runtime-7.x project, we want to gem install rexml
+  # ignoring the dependencies, this is because the pl-ruby version used in these platforms is ancient so it gets
+  # confused when installing rexml. It tries to install rexml's dependency 'strscan' by building native extensions
+  # but fails. We can ignore insalling that since strscan is already shipped with ruby 2 as its default gem.
+  if platform.name =~ /solaris-(10|11)-sparc/ && settings[:ruby_version].to_i < 3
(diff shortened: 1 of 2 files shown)


shubhamshinde360:
Ran impacted projects for all applicable platforms:

agent-runtime-7.x: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3109/

solaris-11-sparc due to xz package version getting updated upstream.
It passes here: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3111/

pe-bolt-server-runtime-main: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3110/

shubhamshinde360:
We will be upgrading to ruby 3.2.5 in main to address the CVEs.

joshcooper:
It looks good to me, though somewhat unrelated to your PR, why does pe-bolt-server-runtime include the rexml gem component? It's a bundled gem in ruby 3.2 so seems unnecessary to add it.

joshcooper:
If we want to Lee pe-bolt-server-runtime as it was before, then I'd recommend allowing the runtime to pass in the old version and default to the newer version for only agent runtime 7.x

donoghuc (approved):
Looking back at this i'm not convinced I actually need rexml in bolt-server. I will go ahead and check that separately this update should be fine, Thanks for the heads up.

joshcooper (approved):

PR #888, opened 2024-08-06
updated packages for bolt release

alex501212:
Updates packages for ruby 3 + puppet 8


configs/components/rubygem-aws-partitions.rb
```diff
@@ -1,6 +1,15 @@
 component "rubygem-aws-partitions" do |pkg, settings, platform|
-  pkg.version "1.960.0"
-  pkg.md5sum "c8b0c649c7a00c79e0af4272debbd8f9"
+  version = settings[:rubygem_aws_partitions_version] || '1.961.0'
+  pkg.version version
+
+  case version
+  when '1.961.0'
```
donoghuc, line 6: Not sure why we would need separate versions for this. This gem is updated very frequenty, perhaps there was an update to it between your gem installs. I believe these versions are compatable with both ruby 2/3

configs/components/rubygem-concurrent-ruby.rb
```diff
@@ -6,6 +6,8 @@
   case version
   when '1.2.3'
     pkg.sha256sum '82fdd3f8a0816e28d513e637bb2b90a45d7b982bdf4f3a0511722d2e495801e2'
+  when '1.3.3'
+    pkg.sha256sum '4f9cd28965c4dcf83ffd3ea7304f9323277be8525819cb18a3b61edcb56a7c6a'
```
donoghuc, line 10: What sets this? Typically concurrent-ruby is a very tricky update. We will need to build this into PE and monitor compatability with jruby etc. I dont see how this is used though currently.

configs/components/rubygem-highline.rb
```diff
@@ -7,6 +7,8 @@
     pkg.md5sum '4209083bda845d47dcc05b7ab23f25fd'
   when '3.0.1'
     pkg.sha256sum 'ca18b218fd581b1fae832f89bfeaf2b34d3a93429c44fd4411042ffce286f009'
+  when '3.1.0'
+    pkg.sha256sum '14183e4628f41fb4211766d0d38fb153f01f1ec39adfa83f5f7c6e6ebeb2b8eb'
```
donoghuc, line 11: Same comment, throughout all these case statements, i'm not seeing how these are used. Perhaps you missed logic in project files? Are all of these due to ruby3/puppet8?

alex501212: Removed the case for 3.0.1 as 3.1.0 is the newest for Ruby 3

configs/components/rubygem-concurrent-ruby.rb
alex501212: I have reverted the changes I made in this file

configs/components/rubygem-aws-partitions.rb
alex501212: Removed the case but just updated to the most recent version

configs/components/rubygem-highline.rb
alex501212: Removed 3.1.0 case

donoghuc:
Not quite following the reverts etc you made. I was pointing out that when you ad a new `case` you must also implement logic to select that case conditionally. One option would be to explicitly call it out in project files, however i'm not convinced that is the best long term solution. Instead i think we should investigate switching on ruby version.

joshcooper:
Before merging, could you squash commits and remove any merge commits from your feature branch?

Also I recommend running this rake task to show how each project in this repo will be modified for a given platform:

```
$ bundle exec rake vanagon:component_diff -- -P all -p el-9-x86_64 --from upstream/master --to HEAD
```

For the most accurate results, you'll want to rebase your feature branch onto `upstream/master` (where `upstream` is the remote for puppetlabs)

donoghuc:
Thanks @joshcooper , i'm going to take over this work. I'll likely just start a new PR.

PR #896, opened 2024-08-14
Bump minitar 1.0.1 for puppet 8

joshcooper:
Minitar 1.x is not backwards compatible with 0.9, so only make this change in agent-runtime-main for now.

Projects that depend on minitar and puppet will need to be made compatible with minitar 1.x before bumping their puppet dependency in puppet-runtime. And when they do bump puppet, they should bump minitar too. I added a reminder for the latter part. This applies to:

    bolt-runtime
    pe-bolt-server-runtime-main
    pe-installer-runtime-main

pdk-runtime depends on minitar, but not puppet. When a new version of puppet is released that depends on minitar 1.x, then modules created using an older version of the pdk, will pull in the latest puppet, but have an incompatible minitar version.

### DIFF
--- a/configs/components/rubygem-minitar.rb
+++ b/configs/components/rubygem-minitar.rb
@@ -1,6 +1,15 @@
 component 'rubygem-minitar' do |pkg, settings, platform|
-  pkg.version '0.9'
-  pkg.md5sum '4ab2c278183c9a83f3ad97066c381d84'
+  version = settings[:rubygem_minitar_version] || '0.9'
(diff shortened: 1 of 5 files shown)


AriaXLi:
Hello, thank you for your contribution! Since this repository is being archived, this PR will be closed. To see more information on this, see [OSPTE-212](https://perforce.atlassian.net/browse/OSPTE-212). If possible, please retarget this PR to the corresponding private internal repository.

PR #898, opened 2024-08-20
(PA-6878) Patch agent-runtime-7.x and main Curl for CVE-2024-7264

amitkarsale:
Patch to fix curl CVE-2024-7264

7.x curl version - 7.88.1
For 7.x patch was inspired from : [http://archive.ubuntu.com/ubuntu/pool/main/c/curl/](http://archive.ubuntu.com/ubuntu/pool/main/c/curl/curl_7.68.0-1ubuntu2.23.debian.tar.xz)

For main - we have the curl 8.9.1 version for which the fix is already present as per  https://curl.se/docs/CVE-2024-7264.html

vanagon-generic-main : https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3215/

vanagon-generic-7.x : https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3214/

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -48,6 +48,7 @@ component 'curl' do |pkg, settings, platform|
     pkg.apply_patch 'resources/patches/curl/CVE-2023-46218.patch'
     pkg.apply_patch 'resources/patches/curl/CVE-2024-2004.patch'
     pkg.apply_patch 'resources/patches/curl/CVE-2024-2398.patch'
+    pkg.apply_patch 'resources/patches/curl/CVE-2024-7264.patch'
   end
 
   configure_options = []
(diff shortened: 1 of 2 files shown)


mhashizume:
This will need to be rebased on main to account for the changes merged in https://github.com/puppetlabs/puppet-runtime/pull/897/

joshcooper:
Could you update the commit message to specify where the patch came from? Also could you mention that this change is only needed in 7.x and not main because curl 8.9.1 already has the fix (according to https://curl.se/docs/CVE-2024-7264.html)

joshcooper (approved):

PR #900, opened 2024-08-22
(PA-6324) Add windowsfips-2016-x64 to agent-runtime-main and agent-runtime-7.x

shubhamshinde360:
- Add the platform file for windowsfips-2016-x64.
 - Add windowsfips-2016 to ruby dtrace for both 7.x and main.
 - Set fips version and fips flags properly.
 - Set NM to the system nm (version 2.42) since mingw nm (version 2.25) conflicts with sed during curl-build.
 - Patch rbconfig.rb to remove the '\r' character appended to CONFIG["ruby_version"] during ruby-build.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -27,6 +27,7 @@ component 'curl' do |pkg, settings, platform|
   elsif platform.is_windows?
     pkg.build_requires "runtime-#{settings[:runtime_project]}"
     pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
+    pkg.environment "NM" , "/usr/bin/nm" if platform.name =~ /windowsfips-2016/
     pkg.environment "CYGWIN", settings[:cygwin]
   elsif platform.is_aix? && platform.name != 'aix-7.1-ppc'
     pkg.environment "PKG_CONFIG_PATH", "/opt/puppetlabs/puppet/lib/pkgconfig"
--- a/configs/components/ruby-2.7.8.rb
+++ b/configs/components/ruby-2.7.8.rb
@@ -141,7 +141,8 @@ component 'ruby-2.7.8' do |pkg, settings, platform|
     'windows-2012r2-x64',
     'windows-2012r2-x86',
     'windows-2019-x64',
-    'windowsfips-2012r2-x64'
+    'windowsfips-2012r2-x64',
+    'windowsfips-2016-x64'
   ]
 
   unless without_dtrace.include? platform.name
@@ -168,6 +169,19 @@ component 'ruby-2.7.8' do |pkg, settings, platform|
         #{settings[:host]} \
         #{special_flags}"
     ]
+  end
+
+  if(platform.name =~ /windowsfips-2016/)
+    # We need the below patch since during ruby build step for windowsfips-2016-x64 agent-runtime builds,
+    # the rbconfig.rb file that gets generated contains '\r' trailing character in 'ruby_version' config.
+    # We patch rbconfig.rb to remove the '\r' character.
+    # This patch has to run after the build step since rbconfig.rb is generated during the build step. 
+    # This is sort of a hacky way to do this. We need to find why the '\r' character gets appended to
+    # 'ruby_version' field in the future so that this patch can be removed - PA-6902.
+    pkg.add_source("#{base}/rbconfig_win.patch")
+    pkg.build do
+      ["TMP=/var/tmp /usr/bin/patch.exe --binary --strip=1 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../rbconfig_win.patch"]
+    end
   end
 
   #########
--- a/configs/components/ruby-3.2.5.rb
+++ b/configs/components/ruby-3.2.5.rb
@@ -161,7 +161,8 @@ component 'ruby-3.2.5' do |pkg, settings, platform|
     'windows-2012r2-x64',
     'windows-2012r2-x86',
     'windows-2019-x64',
-    'windowsfips-2012r2-x64'
+    'windowsfips-2012r2-x64',
+    'windowsfips-2016-x64'
   ]
 
   unless without_dtrace.include? platform.name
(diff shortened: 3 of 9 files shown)


mhashizume:
Could you link to a successful build using these changes?

shubhamshinde360:
Jenkins run:
agent-runtime-7.x: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3195/
agent-runtime-main: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3196/

Artifacts: https://builds.delivery.puppetlabs.net/puppet-runtime/939be69ff93b0455282e8a7914135ed550707162/artifacts/

resources/patches/ruby_27/rbconfig_win.patch
```diff
@@ -0,0 +1,22 @@
...
++  TOPDIR = File.dirname(__FILE__).chomp!("/lib/ruby/2.7.0/x64-mingw32")
+   # DESTDIR on make install.
+   DESTDIR = TOPDIR && TOPDIR[/\A[a-z]:/i] || '' unless defined? DESTDIR
+   # The hash configurations stored.
+@@ -54,7 +54,7 @@ module RbConfig
+   CONFIG["sitedir"] = "$(rubylibprefix)/site_ruby"
+   CONFIG["rubyarchdir"] = "$(rubylibdir)/$(arch)"
+   CONFIG["rubylibdir"] = "$(rubylibprefix)/$(ruby_version)"
+-  CONFIG["ruby_version"] = "2.7.0\r"
++  CONFIG["ruby_version"] = "2.7.0"
```
joshcooper, line 19: I believe `ruby_version` is detected during the ruby build by https://github.com/ruby/ruby/blob/31d0f1a2e7dbfb60731d1f05b868e1d578cda493/configure.ac#L4203 It emits a `conftest.c` program, compiles and runs it, and uses `sed` to extract the version. As mentioned in https://cygwin.com/cygwin-ug-net/using-textbinary.html, cygwin `sed` "Most other programs (such as sed, cmp, tr) use the default mode." The default behavior depends on the filesystem that is mounted. I'm guessing the win2016 cygwin is configured differently than 2012r2.

configs/components/ruby-2.7.8.rb
```diff
@@ -170,6 +171,19 @@
...
+  if(platform.name =~ /windowsfips-2016/)
+    # We need the below patch since during ruby build step for windowsfips-2016-x64 agent-runtime builds,
+    # the rbconfig.rb file that gets generated contains '\r' trailing character in 'ruby_version' config.
+    # We patch rbconfig.rb to remove the '\r' character.
+    # This patch has to run after the build step since rbconfig.rb is generated during the build step. 
+    # This is sort of a hacky way to do this. We need to find why the '\r' character gets appended to
+    # 'ruby_version' field in the future so that this patch can be removed - PA-6902.
+    pkg.add_source("#{base}/rbconfig_win.patch")
+    pkg.build do
+      ["TMP=/var/tmp /usr/bin/patch.exe --binary --strip=1 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../rbconfig_win.patch"]
```
joshcooper, line 183: Why construct the patch command instead of using the built in functionality in vanagon for patching a component?

configs/components/ruby-3.2.5.rb
```diff
@@ -195,6 +196,19 @@
...
+    # We need the below patch since during ruby build step for windowsfips-2016-x64 agent-runtime builds,
+    # the rbconfig.rb file that gets generated contains '\r' trailing character in 'ruby_version' config.
+    # We patch rbconfig.rb to remove the '\r' character.
+    # This patch has to run after the build step since rbconfig.rb is generated during the build step. 
+    # This is sort of a hacky way to do this. We need to find why the '\r' character gets appended to
+    # 'ruby_version' field in the future so that this patch can be removed - PA-6902.
+    pkg.add_source("#{base}/rbconfig_win.patch")
+    pkg.build do
+      ["TMP=/var/tmp /usr/bin/patch.exe --binary --strip=1 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch < ../rbconfig_win.patch"]
+    end
```
joshcooper, line 209: Same question here

configs/components/ruby-2.7.8.rb
shubhamshinde360: Hey @joshcooper,
I think you mean doing
`pkg.apply_patch "#{base}/rbconfig_win.patch"`
instead of the raw command.

I did that earlier but then the apply_patch method adds the patch to 'ruby-patch' step. This fails since we are patching rbconfig file which is generated after the ruby-build step.

joshcooper: Ah right makes sense

configs/components/curl.rb
```diff
@@ -27,6 +27,7 @@
   elsif platform.is_windows?
     pkg.build_requires "runtime-#{settings[:runtime_project]}"
     pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(PATH)"
+    pkg.environment "NM" , "/usr/bin/nm" if platform.name =~ /windowsfips-2016/
```
joshcooper, line 30: I'm not sure what the implications are of using Cygwin `nm` with binaries built by mingw... I assume this problem would go away if we were using a newer version of mingw tool chain?

shubhamshinde360: I was able to find 13.2.0 in the default repository(https://chocolatey.org/api/v2) of choco: https://community.chocolatey.org/packages/mingw
I think this is the latest version. It ships with nm 2.39. Unfortunately, it still results in the `global_symbol_pipe` variable being empty with this version.
The system nm (`/usr/bin/nm`) is version 2.42 which doesn't result the `global_symbol_pipe` issue. So, I manually set it to system nm.

I also tried with mingw 8.1.0 (nm version 2.25) which was available in the default repo, that also fails with the same error.

PR #903, opened 2024-08-23
Updates for Ruby, Openssl, net-ssh, Puppet for Bolt runtime

alex501212:
With these changes I was able to successfully build bolt-runtime for el-9-x86_64

### DIFF
--- a/configs/projects/bolt-runtime.rb
+++ b/configs/projects/bolt-runtime.rb
@@ -3,6 +3,8 @@ project 'bolt-runtime' do |proj|
   proj.setting(:runtime_project, 'bolt')
   proj.setting(:ruby_version, '3.2.5')
   proj.setting(:openssl_version, '3.0')
+  # Legacy algos must be enabled in OpenSSL >= 3.0 for Bolt's WinRM transport to work.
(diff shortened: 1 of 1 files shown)


configs/projects/bolt-runtime.rb
```diff
@@ -1,13 +1,15 @@
...
+  proj.setting(:ruby_version, '3.2.5')
+  proj.setting(:openssl_version, '3.0')
+  # Legacy algos must be enabled in OpenSSL >= 3.0 for Bolt's WinRM transport to work.
+  proj.setting(:use_legacy_openssl_algos, true)
+  proj.setting(:rubygem_net_ssh_version, '7.2.3')
   proj.setting(:augeas_version, '1.14.1')
   # TODO: Can runtime projects use these updated versions?
   proj.setting(:rubygem_deep_merge_version, '1.2.2')
-  proj.setting(:rubygem_puppet_version, '7.32.1')
+  proj.setting(:rubygem_puppet_version, '8.8.1')
```
donoghuc, line 12: puppet is on 8.9.0 now (though we will be doing a release next week) we will want to make sure we are on latest when we ship.

PR #906, opened 2024-08-27
(PA-6896) Enable pie on RHEL 9 and up

skyamgarp:
(PA-6896) remove the shared flag

### DIFF
--- a/configs/components/ruby-3.2.5.rb
+++ b/configs/components/ruby-3.2.5.rb
@@ -94,7 +94,7 @@ component 'ruby-3.2.5' do |pkg, settings, platform|
   special_flags = " --prefix=#{ruby_dir} --with-opt-dir=#{settings[:prefix]} "
 
   if settings[:supports_pie]
-    special_flags += " CFLAGS='#{settings[:cflags]}' LDFLAGS='#{settings[:ldflags]}' CPPFLAGS='#{settings[:cppflags]}' "
+    special_flags += " CFLAGS='-fPIC -fstack-protector-strong -fno-plt -O2' LDFLAGS='#{settings[:ldflags]}' CPPFLAGS='-fPIE #{settings[:cppflags]}' "
   end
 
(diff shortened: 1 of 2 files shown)


configs/components/ruby-3.2.5.rb
```diff
@@ -94,7 +94,7 @@
   special_flags = " --prefix=#{ruby_dir} --with-opt-dir=#{settings[:prefix]} "
 
   if settings[:supports_pie]
-    special_flags += " CFLAGS='#{settings[:cflags]}' LDFLAGS='#{settings[:ldflags]}' CPPFLAGS='#{settings[:cppflags]}' "
+    special_flags += " CFLAGS='-fPIC -fstack-protector-strong -fno-plt -O2' LDFLAGS='#{settings[:ldflags]}' CPPFLAGS='-fPIE #{settings[:cppflags]}' "
```
joshcooper, line 97: Can you describe why you're removing :cflags from here? What problem is being solved aside from "pie doesn't work"? Does this mean none of the `cflags` defined in the project are being passed to the `configure` command?

configs/projects/_shared-compiler-settings.rb
```diff
@@ -12,7 +12,6 @@
 # are too old to support these flags.
 
 if((platform.is_sles? && platform.os_version.to_i >= 15) ||
-      (platform.is_el? && platform.os_version.to_i == 8 && platform.architecture !~ /ppc64/) ||
```
joshcooper, line 15: Does this mean RHEL 9 ppc now supports pie? If so, the commit message should mention the change in behavior.

joshcooper:
Oh sorry I just noticed this is a draft

AriaXLi:
Hello, thank you for your contribution! Since this repository is being archived, this PR will be closed. To see more information on this, see [OSPTE-212](https://perforce.atlassian.net/browse/OSPTE-212). If possible, please retarget this PR to the corresponding private internal repository.

PR #921, opened 2024-10-08
(PA-6962) update curl to address CVE-2024-8096

amitkarsale:
Upgrade Curl to 8.10.1 to mitigate CVE-2024-8096

vanagon generic builder : https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/3296/

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -6,8 +6,8 @@ component 'curl' do |pkg, settings, platform|
   case version
   when '7.88.1'
     pkg.sha256sum 'cdb38b72e36bc5d33d5b8810f8018ece1baa29a8f215b4495e495ded82bbf3c7'
-  when '8.9.1'
-    pkg.sha256sum '291124a007ee5111997825940b3876b3048f7d31e73e9caa681b80fe48b2dcd5'
+  when '8.10.1'
+    pkg.sha256sum 'd15ebab765d793e2e96db090f0e172d127859d78ca6f6391d7eafecfd894bbc0'
   else
     raise "curl version #{version} has not been configured; Cannot continue."
   end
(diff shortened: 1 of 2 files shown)


configs/components/curl.rb
```diff
@@ -53,7 +53,7 @@
   end
 
   configure_options = []
-  configure_options << "--with-ssl=#{settings[:prefix]}"
+  configure_options << "--with-ssl=#{settings[:prefix]} --without-libpsl"
```
cthorn42, line 56: What is the context for adding in `--without-libpsl`? Is this required for every platform we're building on? Does the Curl 7.88.1 package support building with that flag?

amitkarsale: while building, curl 8.10.1 `libpsl`  was expected to be installed which was never the case for prior versions. One was was to add `libpsl` was to either install it as a package while building or bypass using `--without-libpsl`.
I checked with the flag and there were no failures or dependency as such for puppet while building.

I checked building for `curl-7.88.1` with `--without-libpsl` flag and the build was successful. As earlier also there were no such dependency hence it passed.

cthorn42: @amitkarsale we went over this PR in standup and have some suggestions.

Did a bit of research, there is a good article by the developer of Curl on PSL and their use in Curl: https://daniel.haxx.se/blog/2024/01/10/psl-in-curl/

The short of this is PSL support was added a bit ago in the 8.x stream of Curl, previous it used to just warn and looks like in the recent past it started to fail the configure action if the required libraries were not found.

Reading the blog post I think short term it really doesn't change anything to include the flag as is, but we should for sure ticket and make this a known issue.

So if you could make a small change to this PR, pull out the `--without-libpsl` from the commit with the Curl bump and make that its own commit with its own message.

Then would you mind filing a Jira ticket to cover the work of adding libpsl support to puppet-runtime?

amitkarsale: I've created a new ticket to address the libpsl installation. Maybe post release I can take it to unblock the release.
- PA-7099

PR #927, opened 2024-10-31
Update rubygem-r10k to 4.1.0

alex501212:
Updated rubygem-r10k  from 3.16.2 to 4.1.0.

PR #932, opened 2024-11-08
Revert "Update rubygem-r10k to 4.1.0"

alex501212:
Reverts https://github.com/puppetlabs/puppet-runtime/pull/927
Seems to be breaking projects that use r10k


gavindidrichsen (approved):
Cheers @alex501212

PR #935, opened 2024-11-11
Updated rubygem-r10k from 3.16.2 to 4.1.0

alex501212:
Updating to r10k 4 also required the following changes:

- Removed faraday_middleware in bolt-runtime and added faraday-net_http_persistent 2.3.0 as faraday_middleware does not support faraday 2. Further, It is not necessary to use faraday_middleware when using net_http_persistent on Ruby 2.0+ (see - https://github.com/lostisland/faraday_middleware/issues/279#issuecomment-1021023287)
- Updated orchestrator client version from 0.7.0 to 0.7.1 as 0.7.0 had a conflicting faraday version with puppet forge
- Specified puppet forge and faraday/faraday dependency versions in bolt-runtime

Testing:
- Was able to run a passing build for el-9-x86_64 in puppet-runtime and bolt-vanagon
- Bolt acceptance tests passing


gavindidrichsen (approved):
Great work!

